PR #2, opened 2017-05-31
Add back the existing API

chenglou:
This is needed for interop. New rr will be in its separate module.


jordwalke:
Should make things easier to move forward.

PR #6, opened 2017-06-09
Add `create-react-class` to replace deprecated `React.createClass`

davidbarsky:
(I'm moving [this PR](https://github.com/reasonml/reason-react/pull/82) on `reasonml/reason-react` to here. The following text remains the same.)

Hi y'all! This PR resolves issue [#40](https://github.com/reasonml/reason-react/issues/40) by doing two things:

1. Adding [`create-react-class`](https://www.npmjs.com/package/create-react-class) as a peer dependency
2. Changing the BuckleScript external binding from `React.createClass` to `create-react-class` 

Instead of tests (I know, not ideal), I created a [small React app](https://github.com/davidbarsky/create-reason-react-app) that depends on [my fork of reason-react](http://github.com/davidbarsky/reason-react) to demonstrate that everything doesn't immediately blow up.  You can install it by...

```bash
$ git clone git@github.com:davidbarsky/create-reason-react-app.git
$ yarn
$ yarn dev
```

...and opening [http://localhost:8080](http://localhost:8080).

Thanks for the work on reason-react! ❤️


chenglou:
Good stuff. Thanks! But I'm thinking we should model that one as a dependency rather than peerDep? We should soon replace it with our own implementation which isn't too hard: see the class output [here](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&targets=&browsers=&builtIns=false&debug=false&code=class%20asd%20extends%20Component%20%7B%0A%20%20render%3A%201%0A%20%20foo%3A%203%0A%7D)

davidbarsky:
> Good stuff. Thanks! But I'm thinking we should model that one as a dependency rather than peerDep?

Just pushed a change that moves create-react-class as a dependency! I chose to treat `create-react-class` as a peer-dependency because react/react-dom were peer-dependencies, and I followed the example.

> We should soon replace it with our own implementation which isn't too hard: see the class output here.

As for replacing it, yeah, I agree. But to clarify—do you suggest binding BuckleScript to [JavaScript constructors](https://bloomberg.github.io/bucklescript/Manual.html#_binding_to_javascript_constructor_bs_new) or building a separate thing? In regards to the former, my hesitation in using ES6 classes is that [autobinding behavior](https://facebook.github.io/react/docs/react-without-es6.html#autobinding) might change, and I didn't want to risk breaking stuff.

chenglou:
I'm suggesting to take create-react-class's content and make our own version without mixins, getDefaultProps, obj.freeze, and all the other things we don't need. It might likely end up like binding to JS constructors

chenglou:
But this is good, thanks! =D

davidbarsky:
> I'm suggesting to take create-react-class's content and make our own version without mixins, getDefaultProps, obj.freeze, and all the other things we don't need. It might likely end up like binding to JS constructors

Oh, gotcha! I'll start to investigate; curious how this will end up working/good way to spend funemployment...

> But this is good, thanks! =D

Yay! ✨

PR #7, opened 2017-06-09
Add Ruby layout style properties

AgtLucas:


chenglou:
Woah that's a thing... Thanks!

PR #10, opened 2017-06-13
Create index.html

cristianoc:
typo


chenglou:
Thanks. @glennsl's making some syntax checking tool that we should probably use on all our docs. Though here it wouldn't be valid because of `...` anyway.

PR #15, opened 2017-06-15
Add more svg css properties

AgtLucas:
Related to: #14 

Probably we are still missing some svg properties, though.

Issue #14, opened 2017-06-14
Missing `pointerEvents` style attribute


jaredly:

AgtLucas:
@jaredly are you sure? https://github.com/reasonml/reason-react/blob/master/src/reactDOMRe.re#L408 🤔

glennsl:
@AgtLucas That's the prop. SVG CSS properties are largely missing: https://github.com/reasonml/reason-react/blob/master/src/reactDOMRe.re#L915

As are lots of others. It's just a matter of adding them.

AgtLucas:
@glennsl Ah, I see. I'll add then. :)

chenglou:
Do these above 2 solve this?

jaredly:
yup this is solved

PR #17, opened 2017-06-15
Add more svg css properties

AgtLucas:
@chenglou Just pushed more properties here. :)

### DIFF
--- a/src/reactDOMRe.re
+++ b/src/reactDOMRe.re
@@ -912,7 +912,7 @@ module Style = {
     rotation::string? =>
     rotationPoint::string? =>
 
-    /* svg */
+    /* SVG 1.1 - REC */
     alignmentBaseline::string? =>
     baselineShift::string? =>
     clip::string? =>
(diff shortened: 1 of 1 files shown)


chenglou:
kk. Are they complete now? Btw what about https://github.com/reasonml/reason-react/issues/14

AgtLucas:
@chenglou AFAIK is complete now, but we can have a second review.

Which part about the #14?

chenglou:
I'm not too familiar with the pointerEvent. I thought this was related?

src/reactDOMRe.re
```diff
@@ -913,13 +913,33 @@ module Style = {
     rotationPoint::string? =>
 
     /* svg */
```
glennsl: Could you change this to `SVG 1.1 - REC`?

glennsl:
`pointerEvents` was added in #15, so #14 should be good now.

AgtLucas:
@glennsl Done!

chenglou:
Great! Thanks =D

PR #18, opened 2017-06-17
[Suggestion] Add floatToElement

AgtLucas:
First of all, I don't know if this works or not, sorry!

Right now, if we have a `float` value and we wanna show in the DOM, we need to cast types like so:

`ReasonReact.stringToElement(string_of_float myFloatValue);`.

Would be neat if we can use something like `stringToElement` but for float values.

`ReasonReact.floatToElement myFloatValue;`

What do you think?

### DIFF
--- a/src/reactRe.rei
+++ b/src/reactRe.rei
@@ -15,6 +15,8 @@ external stringToElement : string => reactElement = "%identity";
 
 external arrayToElement : array reactElement => reactElement = "%identity";
 
+external floatToElement : float => reactElement = "%identity";
+
 let listToElement : list reactElement => reactElement;
 
 external refToJsObj : reactRef => Js.t {..} = "%identity";


glennsl:
You can also do `ReasonReact.stringToElement {j|$myFloatValue|j}`.

I'm hoping there will eventually be a more ergonomic way of embedding strings in JSX, and better syntax for string interpolation. Combined, those will probably be a nicer way of doing it than adding functions for explicitly converting all the different types.

AgtLucas:
@glennsl Agree!

But are you sure about this syntax?

glennsl:
Pretty sure, but I didn't run it through the compiler. See https://bucklescript.github.io/bucklescript/Manual.html#_unicode_support_with_string_interpolation_since_1_7_0

chenglou:
Yeah @glennsl's solution works. That one casts anything to a string. Uses toString under the hood.
I'll hold off on this PR for now. We might have better solutions coming up soon!

jaredly:
+1 on this -- it would be nice to have `ReasonReact.float`

AgtLucas:
@jaredly I can open a new PR since this one is very old and we don't have the `reactRe.rei` file anymore. :)

rickyvetter:
Hey thanks for adding this and sorry for the delay. Added in #420.

PR #20, opened 2017-06-22
add user-defined shouldUpdate support

rickyvetter:
This adds the actual call to shouldUpdate in the case that a non-silent state update or a props change occurs.

I tried a few ways and I think this is the most readable version of the code. I can rearrange to only call convertPropsIfTheyreFromJs when necessary if we think that function will be expensive, or update with any other readability suggestions.

@chenglou I made a typo while testing this because I expected it to be `::prevState` :P Not sure if it's a breaking change, but I still think it's worth making that move.


chenglou:
> components shouldUpdate doesn't get called on silent updates. if the components shouldUpdate returns false, rerender never happens. if it returns true, rerender happens

PR #24, opened 2017-06-22
[Docs] Add reason-react.css overrides to index.html too

jaredpalmer:
@chenglou Forgot one file :)


chenglou:
Is it possible to have it stop at the position it was spinning toward, when we mouse out?

jaredpalmer:
Looks like it's possible:

https://stackoverflow.com/questions/20052672/pause-animation-on-hover

PR #26, opened 2017-06-30
Fix shouldComponentUpdate comment position

istarkov:
It was placed above `componentWillReceiveProps`


chenglou:
Thanks! cc @rickyvetter

PR #31, opened 2017-07-09
[Docs] Move event.target.value example

bjrnt:
I thought it would be good to move the `event.target.value` example up to the `Working with DOM` section, as it was pretty hard to find previously.

When trying to figure out how to access `event.target.value` most people's flows (like mine) are probably something like this:
1. CMD+F "event"
2. See the `If you’re accessing fields on your event object...` text and click the link.
3. Read the `That’s all you need.` part and be stumped.
4. Give up and search discourse chat history.

Moving this example up a bit would help it catch the reader's eye.


chenglou:
Good point, thanks =)

PR #37, opened 2017-07-17
Add loop attribute

logason:
Add `loop` attribute for `<video />`


chenglou:
There sure are a lot of these...
Thanks!

PR #44, opened 2017-08-06
Add more supported React attributes

anmonteiro:


chenglou:
Thanks! cc @rickyvetter when we do the internal sync

PR #48, opened 2017-08-12
Kick broken crra out in favor of reason-scripts

glennsl:
I could also barely find the getting started guide. It's pretty well hidden at the top there.


chenglou:
Thanks!

PR #49, opened 2017-08-14
Fixed React.cloneElement bs.module casing

bloodyowl:
This caused a "There are multiple modules with names that only differ in
casing" warning when using webpack to bundle BuckleScript output


chenglou:
Awww crap syncing error from internal version. Thanks!

PR #50, opened 2017-08-15
Prefix begin prop with underscore

logason:
`begin` is a [reserved keyword in ocaml](http://caml.inria.fr/pub/docs/manual-ocaml-4.00/manual044.html) so we need to rename the `begin` prop to `_begin`


kennetpostigo:
@chenglou @cristianoc not sure if `super_errors` has good messages for these types of errors. Should there be?

chenglou:
It's a syntax error, so belongs to Reason

PR #51, opened 2017-08-16
Update link to reasonml module documentation

lpalmes:
Updated the link to `Pass in Components Class as a Prop` explanation.
Also might set the record for smallest pr in the history.


chenglou:
Thanks! =D

PR #56, opened 2017-09-01
Fix `this` bug in Async State Setting JS example

af:
Classic ES5 footgun :)


chenglou:
One of the few snippets of JS in there and I had to commit a bug, sigh

Thanks!

PR #69, opened 2017-09-26
update React version constraint

IwanKaramazow:
Reason-React installs React 15 in `node_modules/reason-react/node_modules/react`.
When working with fiber in `node_modules/react` two Reacts got bundled, which resulted in a few errors.


chenglou:
Thanks!

PR #70, opened 2017-09-27
Docusaurus

rickyvetter:
Latest demo: https://reason-react-bsqywzdcms.now.sh/

New docs site. All static (no JS) and very fast. Opened up into multiple pages. Content still needs work, but that can be done iteratively. Would love feedback on what can be improved.


bassjacob:
Having a click around, looks really nice so far. Is this the last planned transition for the docs?

bassjacob:
Also, is the source code for `docusaurus` available anywhere? I couldn't see it on gh, nor under the author's name anywhere, only the npmjs page for it.

rickyvetter:
Yes it is the last planned. Gatsby is having trouble scaling to the number of pages we'd have with translation, and we run into some other issues with the complex webpack toolchain. I'm also trying phenomic and working to see what we can fix with gatsby, but so far Docusaurus has been the most impressive in initial setup and simplicity. It seems to handle translations very nicely.

That said, this isn't a definite solution. If we can make gatsby work I'd rather avoid the churn, but there are some big pain points.

chenglou:
All the inter-docs links are broken. We'll need to fix them before publishing. So let's keep the new docs under `docs2` and old docs still at `docs`.

We should also manually stub each broken link, like I did here: http://facebook.github.io/reason/gettingStarted.html. I can do that too. This way when we switch over, all the existing links will have a helpful error. It doesn't create a great impression if someone's conference talk links are stale a few days after posting them (I know a few people are preparing ReasonReact talks right _now_). Same for all the reddit/SO rr front-page links.

Docsearch working? We need it for the release since the docs are multi-page now.

Let's also make sure pjax is solid and lightweight.

kennetpostigo:
@rickyvetter would it be possible to make a general base documentation repo for future reason related libraries? I currently have a custom one for reason-node, but I would love to be able to use a base one like this

chenglou:
@kennetpostigo that's already the case. Stay tuned

rickyvetter:
@chenglou I've corrected all of the broken internal links and added redirects for examples, faq, getting started, blog, and index (only if hitting a hash that existed). Not sure I follow on docs vs docs2 - once the site is generated the locations in source don't matter.

PR #71, opened 2017-09-27
Remove enqueue from the API

cristianoc:
@chenglou should we remove enqueue?
It's currently broken as it takes state instead of self in the implementation, while the .rei specifies self.


chenglou:
I'll remove it in the next next version

PR #85, opened 2017-10-05
Clarify goals and differences

canadaduane:
For a newcomer like myself, more clarity around differences between `Bsb` and `Reason-scripts` would have helped me understand what the trade-offs were, and that a server is not needed or necessary in the `Bsb` case.


chenglou:
Thanks! Moved one sentence.

PR #86, opened 2017-10-05
minor wording change to creation docs

thatbraxguy:
I noticed some awkward wording while reading through the docs. super minor 🐙


chenglou:
Thanks. Should we just say "...asks you to return a component record, the one you created above"?

thatbraxguy:
Going in that direction would "asks you to return the component record created above" work?

chenglou:
Yep

thatbraxguy:
updated

chenglou:
Thanks!

PR #87, opened 2017-10-06
Improve CSS animation

kutyel:
Hi!

I ❤️  the idea of making the React logo spin on the background, but the current animation is almost imperceptible, so I decreased the animation time and you can notice now the movement, which has a nice transition without distracting the user from the actual docs.

Awesome work both with Reason and ReasonReact btw!

cc: @jordwalke 😉

### DIFF
--- a/website/static/css/custom.css
+++ b/website/static/css/custom.css
@@ -100,7 +100,7 @@ a.hash-link {
   width: 800px;
   left: 0px;
   bottom: -400px;
-  animation:spin 750s linear infinite;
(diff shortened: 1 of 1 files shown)


rickyvetter:
The imperceptibility is my favorite part of this animation. It makes you do a double-take to check if it actually was moving. I'd rather keep it as is.

chenglou:
Yeah let's keep it subtle... It'll be an Easter egg for those who stare at the page long enough!

Closing. Thanks for the contribution though. Keep them coming =)

PR #90, opened 2017-10-08
[Docs] Fix mobile navigation styles

imagentleman:
On mobile devices narrower than 375px, the search navigation and logo overlap a little.

A 320px device currently looks like this:

current-320

And my phone which is 360px looks like this:

current-360

Also, the "header title" and the search input are aligned by the baseline (at least on 375):

current-375

So, this PR drops the search between the logo and the rest of the links on its own row, when the viewport is below 375 and moves the search a little bit down to center align it (i think it would look better-ish...).

It would look like this on 320px:

new-320

And like this on 375px:

new-375

The dropdown goes below the search as expected and would look like:

dropdown

PS:
* Devices below 375px account for about 42% of mobile devices (http://gs.statcounter.com/screen-resolution-stats/mobile/worldwide), so i think this fix or something similar would be worth it.
* Docusaurus' (https://docusaurus.io/) repo (https://github.com/facebookexperimental/docusaurus) seems to be private at the moment so i couldn't send the PR there.

### DIFF
--- a/website/static/css/custom.css
+++ b/website/static/css/custom.css
@@ -228,11 +228,21 @@ code {
 .navSearchWrapper {
   align-self: center;
   position: absolute;
-  top: 15px;
+  top: 55px;
   right: 10px;
+  left: 10px;
   display: flex;
   justify-content: center;
   align-items: center;
+  width: auto;
+}
+
+@media only screen and (min-width: 375px) {
+  .navSearchWrapper {
+    top: 16px;
+    left: auto;
+    width: 50%;
+  }
 }
 
 @media only screen and (min-width: 1024px) {
@@ -285,6 +295,7 @@ code {
 }
 
 .navigationSlider .slidingNav {
+  background-color: $primaryColor;
   /* fixed makes this calculation *much* easier */
   box-sizing: border-box;
   position: fixed;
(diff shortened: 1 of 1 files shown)


chenglou:
cc @JoelMarcey, @ericnakagawa

@rickyvetter what do you think? Or should directly drop to 3-lines (search on its own line) for all mobile view?

rickyvetter:
This looks great to me! I'd rather keep in on the right as long as possible because the header is sticky. Three bars take up quite a bit of vertical space. In fact, I think it'd be worth investigating unsticking the header when we drop to three.

chenglou:
We can unstick only the reason logo part and keep the last two lines

imagentleman:
Now that i think about it, having three sticky lines does feel a little excessive. What about just dropping the "ReasonReact" text on that specific range (320 - 375), or something along those lines (keepin the text but removing the logo, etc)...

This:

reasonml github io-reason-react- iphone 5

Or this:
reasonml github io-reason-react- iphone 5 1

chenglou:
I'm fine with either the current PR, or just keeping the `ReasonReact` text. Just the logo feels a bit empty.

JoelMarcey:
Thanks @chenglou. I have a conversation going offline with @rickyvetter about this.

rickyvetter:
Closing in favor of #94. Thanks for working through this with us.

PR #91, opened 2017-10-10
Point to JSX desugaring from the children-as-props special case docs

jasim:
JSX is usually sufficient to write ReasonReact except when needing to pass
children as props. This is the one of the few times people need to understand
how JSX is de-sugared in Reason, and so a pointer to JSX docs is necessary
when getting stuck.


chenglou:
Thanks! Btw, `jsx.md#capitalized` is a legit link that's parsed and transformed to the corresponding URL. Works with i18n URLs too =)

PR #94, opened 2017-10-12
 [Docs] Fix mobile navigation styles

imagentleman:
Based on the problem and comments from here: https://github.com/reasonml/reason-react/pull/90

Went ahead and added this pr that implements the second suggestion from this comment: https://github.com/reasonml/reason-react/pull/90#issuecomment-335337753

The Nav would look like this (btw, the screenshot was taken with firefox responsive design mode):

screen shot 2017-10-11 at 21 13 05

### DIFF
--- a/website/static/css/custom.css
+++ b/website/static/css/custom.css
@@ -9,6 +9,16 @@
   transform: rotate(1080deg);
 }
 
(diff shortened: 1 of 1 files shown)


rickyvetter:
This looks great! Thanks.

PR #103, opened 2017-10-24
Add an 2048 example project

LIU9293:
I am a react user now learning reason by complete this small project and translating reason react doc into Chinese. I will be very happy to see my project on the doc, thanks.


chenglou:
Sorry for the late reply. Yes, thanks so much for translating the docs!

PR #110, opened 2017-10-30
[docs]: [syntax 2 to 3] Update creation-props-self.md

OlegLustenko:
https://reasonml.github.io/reason-react/docs/en/creation-props-self.html


chenglou:
Merged all your diffs. You just saved me _so_ much time, thank you!

PR #114, opened 2017-11-02
Fix typos in state-actions-reducer.md

ellbee:
Fix a couple of typos in the first example.

This example still won't compile because ReducerComponent requires an action to be used in the reducer function. Maybe it would be best to remove it?


chenglou:
Thanks!

PR #117, opened 2017-11-06
Svg attribute fix

kennetpostigo:
move `_` to the front of the attribute


kennetpostigo:
cc @chenglou

chenglou:
Just realized I fixed this internally but haven't synced it out

PR #119, opened 2017-11-07
Update bsconfig.json: explicitly set reason version to 2 (until we upgrade to reason 3)

viebel:


remithomas:
@viebel you should remove the previous tag [line 10](https://github.com/reasonml/reason-react/pull/119/files#diff-f41ca12831738be5112886ab27935569R10)

chenglou:
Thanks! I'm making a small release soon so I'll change this to 3 here.

PR #120, opened 2017-11-08
[Docs] Update to reason 3 instance-variables.md

InsidersByte:
The `More details here` link errors when clicking on it with a Syntax error as it's using the reason 2 syntax


chenglou:
Aha, nice catch, thanks.

PR #121, opened 2017-11-08
Fix typo in state-action-reducer.md

jaisonjustus:
In code the type is defined as `action`. But in the explanation its written as `There's a user-defined type called actions`. Changing actions to action.


chenglou:
Thanks!

Issue #122, opened 2017-11-10
Kill ReasonReact.createDomElement


chenglou:

peterpme:
For the sake of cleaning up the repo (and given how old this issue is) I'm going to close this out.

Please re-open if this is still relevant. Thanks!

PR #126, opened 2017-11-17
Remove ReasonReact.createDomElement.

jdhorwitz:
Fixes issue https://github.com/reasonml/reason-react/issues/122

Removed ReasonReact.createDomElement.

### DIFF
--- a/docs/children.md
+++ b/docs/children.md
@@ -79,18 +79,3 @@ Here are some use-cases for children + children spread + Reason built-in data st
 How do we know that `MyForm`, `Motion` and `Layout` accept such children? Well, that'll simply be inferred by internal usage of `children` in these components' respective `render`. Just another day in the magical land of type inference.
 
 Go wild!
-
-### Pitfall
-
-Following the above section's reasoning, this code should work:
-
-```reason
-let children = [| <div /> |];
-<div> ...children </div>; /* <--- this line */
-```
-
-It _does_ work for custom components (upper-case), But it doesn't for DOM components (lower-cased). This is due to some bindings-related constraints; we'll make this better in the future! In the meantime, for passing to DOM element, e.g. `<div className=bar> ...children </div>` use the following:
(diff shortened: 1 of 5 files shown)


chenglou:
Sorry, I should have explained myself more clearly in the issue. `createDomElement` should be removed, but we need to fulfill its use-case better, otherwise we're literally removing an escape hatch =).

It'll probably be removed in the next next release, or so.

cullophid:
@chenglou I assume the issue is `<div> ...children </div>` ?

chenglou:
Lower-case desugared into a call that uses bs.splice, so they need an array literal. We did this out of care for perf. A few extra calls for each div didn't sound nice. I might have a few ways to circumvent these.

cullophid:
can it not be handled by the JSX preprocessor?

chenglou:
It can. We'll likely do that. Just that it's quite a bit of stuff in the JSX already. But yeah

cullophid:
@chenglou true but then you dont have to do it at runtime.

tmepple:
I've worked around this by using React fragments.  First by defining the external:

```
module Fragment = {
  [@bs.module "react"] external reactClass : ReasonReact.reactClass = "Fragment";
  let make = children =>
    ReasonReact.wrapJsForReason(~reactClass, ~props=Js.Obj.empty(), children);
};
```

Then just `<Fragment> ...children </Fragment>`

I'm not sure if it's as efficient as `createDomElement` or if there are other downsides but it works well without creating an extra `<div>`!

peterpme:
Hey @jdhorwitz in an effort to clean up the repo, are you ok with closing this out and re-opening if its still valid?

rickyvetter:
I don't think that this is relevant for the new JSX. This function will get deprecated with the rest of ReasonReact. Thanks for the discussion and sorry for the slow movement here.

PR #127, opened 2017-11-17
Fix broken links in children blog entry to docs

rafayepes:


chenglou:
@rickyvetter the blog posts are a bit more special-cased than I thought! Or rather, less special-cased.

Thanks @rafayepes!

PR #136, opened 2017-12-05
Update jsx.md

kpaxqin:
Component name should not be changed after translate.

### DIFF
--- a/docs/jsx.md
+++ b/docs/jsx.md
@@ -70,14 +70,14 @@ The `make` above is exactly the `make` function you've seen in the previous sect
 ReasonReact children are **fully typed**, and you can pass any data structure to it (as long as the receiver component permits it). When you write:
 
 ```reason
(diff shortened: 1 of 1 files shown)


chenglou:
Sorry not sure what you mean; why not `MyForm`?

kpaxqin:
@chenglou 
Either one is okay but in [line: 88](https://github.com/reasonml/reason-react/pull/136/commits/a887a17647ec018a30fd5fdf2866fd08b402999e#diff-aeaebab7360181a06c66ed833b77ba66R88) we say it will be translated to `MyReasonComponent`, and same as line 104. 

I believe its a typo, component name wouldn't be changed after translate from my understanding. This PR just change all `MyForm` to `MyReasonComponent` for consistency.

chenglou:
Aaah you're right, thanks for catching it!

PR #142, opened 2017-12-15
Svg prop

kennetpostigo:
adds a svg property

### DIFF
--- a/src/ReactDOMRe.re
+++ b/src/ReactDOMRe.re
@@ -67,6 +67,7 @@ external props :
     ~className: string=?, /* substitute for "class" */
     ~contentEditable: Js.boolean=?,
     ~contextMenu: string=?,
+    ~dataName: string=?
     ~dir: string=?, /* "ltr", "rtl" or "auto" */
     ~draggable: Js.boolean=?,
     ~hidden: Js.boolean=?,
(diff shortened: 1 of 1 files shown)


src/ReactDOMRe.re
```diff
@@ -67,6 +67,7 @@ external props :
     ~className: string=?, /* substitute for "class" */
     ~contentEditable: Js.boolean=?,
     ~contextMenu: string=?,
+    ~dataName: string=?
```
AgtLucas: Missing a `,`

kennetpostigo: thanks!

AgtLucas: :)

chenglou:
Thanks!

chenglou:
Hold on, where is this attribute? https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes

mobily:
hm 🤔 it's not valid SVG attribute

https://www.w3.org/TR/SVG/attindex.html
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute

kennetpostigo:
My bad, not sure why sketch adds this when generating an SVG, when removed the svg renders properly

chenglou:
Some random metadata I guess! Closing =)

PR #145, opened 2017-12-21
fix command "npm run webpack" in homepage

clouds56:
it is not consistent with https://reasonml.github.io/reason-react/docs/en/installation.html


chenglou:
Thanks!

PR #146, opened 2017-12-25
Small creation-props-self.md copy fix

riwsky:


chenglou:
Thanks. I've changed it to just "language" because it's available in ocaml too as a function feature

PR #147, opened 2017-12-25
Improve "Component as Prop" doc

MoOx:
Following a discussion on Discord, here my 2cents to improve this docs. You may want to adjust this, that's probably just a start :)

### DIFF
--- a/docs/component-as-prop.md
+++ b/docs/component-as-prop.md
@@ -15,3 +15,35 @@ This also has some added advantages:
 
 - The _owner_ is the one controlling how the component renders, not some opaque logic inside `MyBanner`.
 - Circumvents the tendency of needing a [props spread](props-spread.md) in `MyBanner`'s render, since passing a component to a child and the child using prop spread on said component usually go together.
+
+--- 
+
+In practise, if you want a component that would accept children, you can do the following:
+
+```reason
+<RandomComponent wrapperFunc=((~children) => <div className="something-special"> ...children </div>) />
+```
+
+The associated definition of the `RandomComponent` could be something like:
+
+```reason
+let component = ReasonReact.statelessComponent("RandomComponent");
+
+let make =
+    (
+      /* note the default value that just wrap with a simple node */
+      ~wrapperFunc=(~children) => <div> ...children </div>,
+      _children
+    ) => {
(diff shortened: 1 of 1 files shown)


docs/component-as-prop.md
```diff
@@ -15,3 +15,35 @@ This also has some added advantages:
 
 - The _owner_ is the one controlling how the component renders, not some opaque logic inside `MyBanner`.
 - Circumvents the tendency of needing a [props spread](props-spread.md) in `MyBanner`'s render, since passing a component to a child and the child using prop spread on said component usually go together.
+
+--- 
+
+In practise, if you want a component that would accept children, you can do the following:
```
vnctaing: hey ! there's a small typo here "In pratise" > "in practice"

bloodyowl: `if you want a component that accepts children`. + I'd invert the code samples and put `then you can use it like this:` between the two

MoOx:
@vnctaing @bloodyowl adjusted!

jaredly:
Way cool, thanks for getting this started! Would this simpler example be sufficient?

```
If you want to "pass in" a component that takes children (including dom elements like "div", etc.) you could do:

```reason
<RandomComponent wrapperFunc=((~children) =>
    <SomeComponentThatTakesChildren>
     ...children
    </SomeComponentThatTakesChildren>
) />
\```
```

MoOx:
If that's the only content you want to put instead of this PR, I would find this a bit light, especially for new comers from reactjs.
Maybe something in the middle can en enough (eg: no defaults as in the example) but I would prefer to see something more concrete than an incomplete example that would require more works to have it working (again I am speaking as someone used to do this kind of things in reactjs that struggled to find the right approach in reason).

jaredly:
Yeah, that makes sense :) if you remove the default, then I think it's good to go!

peterpme:
Hey @MoOx,

In an effort to clean this shizzle up, do you mind closing this out and re-opening if its still relevant? I _promise_ we'll be able to act on it faster :smile:

Thanks & sorry about the delay!!

peterpme:
Closing this out for now since its outdated docs :smile:

PR #151, opened 2018-01-02
Adding Fragment class

ccutch:
Adding Fragment class that uses `React.Fragment` from https://reactjs.org/docs/fragments.html. Need to add interface definition and compatibility check with older versions of react.

### DIFF
--- a/src/ReasonReact.re
+++ b/src/ReasonReact.re
@@ -754,3 +754,10 @@ module WrapProps = {
 };
 
 let wrapJsForReason = WrapProps.wrapJsForReason;
+
+/* reference: https://reactjs.org/docs/fragments.html */
+module Fragment = {
+  [@bs.module "react"] external fragment : ReasonReact.reactClass = "Fragment";
+  let make = (children) =>
+    ReasonReact.wrapJsForReason(~reactClass=fragment, ~props=Js.Obj.empty(), children);
+};


jaredly:
Yeah, I would love to get this in! Any ideas about compatability checking with older versions of react? We probably just have to do a breaking version update that requires react 16

d4h0:
Hi,

Any idea when Fragment will be included?

Here is an additional benefit of Fragment for reason-react apps:

At the moment it is not possible to use the child spread syntaxt with lower case HTML components:

`<MyComponent> ...children </MyComponent>` - this works

`<div> ...children </div>`- this doesn't work.

Instead the [documentation](https://reasonml.github.io/reason-react/docs/en/children) says, we should use the following:

```
ReasonReact.createDomElement("div", ~props={"className": bar}, children)
```

with Fragment (at least with the implementation from this PR) we could do the following:

```
<div>
  <Fragment>
    ...children
  </Fragment>
</div>
```

which I'd prefer.

Another benefit:

the following seems to be not valide (gives me an unknown syntax error):

```
<MyComponent>
  <h1> {ReasonReact.string("foo")} </h1>
  ...children
</MyComponent>
```
With Fragment we can workarround this without creating an additional HTML element:

```
<MyComponent>
  <h1> {ReasonReact.string("foo")} </h1>
  <Fragment> ...children </Fragment>
</MyComponent>
```

I have added Fragment to my project, but an official implementation would be really awesome.

Even more awesome would be support for the [short <></> syntax](https://reactjs.org/docs/fragments.html#short-syntax) ;)

chenglou:
I've provided an alternative solution to Fragment which doesn't change how ReactJS sees it. It has a few limitations we may be able to work out in the future. But it works for most of the use-cases.

Sorry for the delay on this, and thanks again!

naartjie:
Referencing the docs for [Fragment](https://reasonml.github.io/reason-react/docs/en/jsx#fragment) in case anyone else lands on this issue.

PR #153, opened 2018-01-04
[Docs] Fix error in self section

rubencosta:
Since `self` is not passed to `make` I assumed it would be `render`.


chenglou:
Thanks!

PR #156, opened 2018-01-10
Fix minor router typo

pbomb:
The example url record had a typo. It named the `search` field as `path`.


chenglou:
Typical copy paste... thanks!

PR #159, opened 2018-01-10
[docs] Remove duplicated sentence from "Working with Dom" page

hmillison:
I noticed that there is an unexpected sentence at the start of this page of the documentation. Since it's repeated in the subsequent paragraph, i'm assuming that it was an unintentional duplicate.

Love the work you're all doing with ReasonML!


chenglou:
Thanks!

PR #161, opened 2018-01-10
Add switch parens

thangngoc89:
I think that the parens were removed by mistake on this commit https://github.com/reasonml/reason-react/commit/eaa4887c7d7352b20f67863c724d5646e0ef38f0


chenglou:
oops, thanks.

PR #183, opened 2018-02-18
Expose parser for path and search

ulrikstrid:
This is to help with server side rendering. Example usage would be in this example https://github.com/jaredpalmer/razzle/pull/513.

Opening this PR to move a discussion we had on Discord to here on Github.


ulrikstrid:
I guess this is irrelevant after this much time, closing

PR #191, opened 2018-02-25
Move to send api in ReactMini

rickyvetter:
This matches the React api and allows for a little more consistent examination of the codebase.

cc @cristianoc


jordwalke:
Awesome thanks.

PR #201, opened 2018-03-15
Fix router incompatibility with IE11.

cknitt:
`element.dispatchEvent(new Event(eventName))` won't work in IE11, see https://stackoverflow.com/questions/27176983/dispatchevent-not-working-in-ie11.

This PR provides a safe `makeEvent` function that works both in modern browsers and in IE11.


cknitt:
@chenglou PR is updated. I got rid of the popState polymorphic variant, and JS looks much better now.

chenglou:
Thanks!

PR #203, opened 2018-03-18
Fix "In the same file" for docs/simple.md example

jeanbaptisteassouad:
Add an "ReasonReact.element" to make it work in the same file


chenglou:
Oops... thanks!

PR #205, opened 2018-03-27
remove silent update

rickyvetter:
This removes SilentUpdate as discussed a few times before. Right now, React does not provide an API suitable to build out SilentUpdate without bookkeeping and implementing a custom sCU. This isn't a great solution and it might be good to work with the React team to find something that doesn't require bookkeeping on our end.

Pros:
* Removes a lot of complexity in RR and all custom lifecycle bookkeeping.
* SilentUpdate was hard to explain to newcomers
* As implemented, SilentUpdate is not Fiber async safe. Fiber can rollback so you can't use SilentUpdate to store side-effectual stuff like timer ids, DOM refs.
* It isn't documented, so breaking change will not be widespread. In theory codemodding to Update will just be a perf loss (but seems unlikely).
Cons:
* Breaking change.
* For some specific performance cases this API is desirable.

### DIFF
--- a/lib/js/src/ReasonReact.js
+++ b/lib/js/src/ReasonReact.js
@@ -128,45 +128,19 @@ function createClass(debugName) {
                           return /* tuple */[
                                   /* None */0,
                                   {
-                                    reasonState: reasonStateUpdate[0],
-                                    reasonStateVersion: curTotalState.reasonStateVersion + 1 | 0,
-                                    reasonStateVersionUsedToComputeSubelements: curTotalState.reasonStateVersionUsedToComputeSubelements
+                                    reasonState: reasonStateUpdate[0]
                                   }
                                 ];
                       case 1 : 
                           return /* tuple */[
-                                  /* None */0,
-                                  {
-                                    reasonState: reasonStateUpdate[0],
-                                    reasonStateVersion: curTotalState.reasonStateVersion + 1 | 0,
-                                    reasonStateVersionUsedToComputeSubelements: curTotalState.reasonStateVersionUsedToComputeSubelements + 1 | 0
-                                  }
+                                  /* Some */[reasonStateUpdate[0]],
+                                  curTotalState
                                 ];
                       case 2 : 
                           return /* tuple */[
-                                  /* Some */[reasonStateUpdate[0]],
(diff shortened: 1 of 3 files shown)


flash-gordon:
> This removes SilentUpdate as discussed a few times before

@rickyvetter cool, is there a comprehensive list of features yet to be removed? :) I can adjust my code easily in this case but I didn't expect I would need to. I searched for "SilentUpdate" on the discord channel and didn't find any mentions it's gonna be removed. I'm not complaining, just wondering if I'm missing something.

rickyvetter:
@flash-gordon good question! SilentUpdate is something we've thought about removing for a while now and intentionally didn't advertise or document to limit any potential issues. It was originally planned as a way to update refs/simulate instance variables but we've actually discovered some issues with this in the async world in discussions with the React team. (Those are the discussions I'm referring to.) Now they only have use as a perf optimization and IMO that doesn't justify the extra bookkeeping and complexity on every component since it's not clear that the perf would outweigh the cost.

There is a larger breaking change coming inline with the [changes coming to React](https://github.com/reactjs/rfcs/blob/master/text/0006-static-lifecycle-methods.md) but the exact way we do this change hasn't been figured out yet. We will almost certainly ship with a codemod that will minimize changes on your end when possible.

All of these changes should be listed in https://github.com/reasonml/reason-react/blob/master/FUTURE.md and they are not. That's my bad. I'll be sure to update that with my current plan.

flash-gordon:
> All of these changes should be listed in FUTURE.md and they are not. That's my bad. I'll be sure to update that with my current plan.

Thanks, will keep an eye on it! I know maintenance can be a burden but keeping things published saves time in the long run.

chenglou:
`SilentUpdate` and `SilentUpdateWithSideEffect` should never be used in userland; this is described in the ReasonReact docs. If you're using it for e.g. updating an instance variable then it's a bug; you should be using a ref cell instead: https://reasonml.github.io/reason-react/docs/en/state-actions-reducer.html#state-update-through-reducer

flash-gordon:
@chenglou yeah, this is most likely a bug, those places should be updated with `register` actually, once it's there

Issue #181, opened 2018-02-17
Docs - Change SVG favicon to PNG


markus-willems:
Is it possible to change the favicon from SVG to PNG (as it is for the normal [Reason Docs](https://reasonml.github.io/))?

Currently only Firefox (fully) supports the use of SVG as favicon: https://caniuse.com/#feat=link-icon-svg

In my bookmarks bar I only display the favicons (no text) and that standard Chromium icon really messes with my OCD :smile:

```html
<link rel="shortcut icon" href="/reason-react/img/reason-react-red.svg"/>
```

matejstrasek:
This can be closed as https://github.com/reasonml/reason-react/pull/207 was merged.

chenglou:
Thanks!

PR #207, opened 2018-03-30
Change favicon to png

matejstrasek:
As mentioned in https://github.com/reasonml/reason-react/issues/181 using SVG as favicon doesn't have good cross-browser support (https://caniuse.com/#feat=link-icon-svg).

- Works on Firefox:

reason-react-success-firefox

- Fails on Chrome:

reason-react-fail-chrome

With this pull request favicon is changed from SVG to PNG.

- Works on Firefox:
reason-react-fix-firefox

- Works on Chrome:
reason-react-fix-chrome


chenglou:
Thanks

PR #217, opened 2018-04-21
Typo - use javascript types rather than string

JasoonS:
Make values inside `div` actually of different types rather than just a string of "1" or "null" respectively.


chenglou:
Thanks

PR #224, opened 2018-05-13
Prevent warning in intro-example, add example of usage

myshov:
- Removed potential warning of unused `name` variable
- Added explicit sections with component definition and its usage


jaredly:
awesome, thanks!

PR #226, opened 2018-05-16
Adding Context

Hehk:
This is an initial attempt to bring the context API to reason-react. This version is implemented in pure Reason and does support working with react 15.

If React 15 support is no longer a priority I believe this can be done through a simple external on the ReactJS context API.

This code is a copy of a library I wrote: https://github.com/Hehk/reason-react-context

### DIFF
--- a/lib/js/src/ReasonReact.js
+++ b/lib/js/src/ReasonReact.js
@@ -1,6 +1,5 @@
 'use strict';
 
-var Block = require("bs-platform/lib/js/block.js");
 var Curry = require("bs-platform/lib/js/curry.js");
 var React = require("react");
 var Caml_builtin_exceptions = require("bs-platform/lib/js/caml_builtin_exceptions.js");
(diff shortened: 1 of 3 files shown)


Hehk:
`
[@bs.module "react"] external createContext : 'a => Js.Dict.t(ReasonReact.reactClass) = "";

module type ContextConfig = {type t; let debugName: string; let value: t;};

module CreateContext = (C: ContextConfig) => {
  let context = createContext(C.value);
  module Provider = {
    let make = (~value: C.t, children) =>
      ReasonReact.wrapJsForReason(
        ~reactClass=Js.Dict.unsafeGet(context, "Provider"),
        ~props={"value": value},
        children,
      );
  };
  module Consumer = {
    let make = (children: C.t => 'a) =>
      ReasonReact.wrapJsForReason(
        ~reactClass=Js.Dict.unsafeGet(context, "Consumer"),
        ~props=Js.Obj.empty(),
        children,
      );
  };
};
`

I messed around locally and this seems to work as an external (I am not the best at doing external so their might be a simpler way X.x)

jaredly:
Thanks! This pure reason implementation has some behavioral differences to the react context api, which might be confusing. I'd probably vote to use the external to react 16, if we're willing to drop support for 15 (cc @chenglou). Otherwise, we could modify this PR to more closely match the react 16 behavior.

chenglou:
Let's not reimplement it. My heuristics is "am I able to maintain this" lol

We'll drop 15 later

jaredly:
Here's a simple wrapper of the react 16 context API if anyone wants it: https://gist.github.com/jaredly/951177b8f5518bfbd19aa1a5ce093e7b

sslotsky:
@Hehk @jaredly have you thought about what updating should look like if your consumer is nested? In the [React docs](https://reactjs.org/docs/context.html#updating-context-from-a-nested-component) they create a component that wraps the provider, attaches update functions to the state, and passes in the state as the provider's value. If you wanted to do something similar, how would you do it? Maybe pass in a tuple containing the state and a record of update functions that delegate to `self.send`? How would something like that perform in terms of re-rendering?

jaredly:
@sslotsky yup, you could basically use the same pattern they do

sslotsky:
@jaredly Sure, except I don't think you can really attach the update handlers directly to the `state` object as they do in the example (or can you? Am I thinking about this wrong?) So it seems like you either have to provide a tuple or use some other form of composition. I'm wondering how that performs, because as I understand it, attaching the update handlers to the `state` object is done in order to retain referential identity for the value that gets passed to the provider, so that excessive re-rendering is avoided. Do I have that wrong?

daniloraisi:
Using the same Context Binding from @jaredly, I created another example with complex context type for a themed bootstrap button: [https://gist.github.com/daniloraisi/6f76691c7f1d3d67f73a8506d34858f8](https://gist.github.com/daniloraisi/6f76691c7f1d3d67f73a8506d34858f8)

You can check a working Codesandbox too: [Reason Context API](https://codesandbox.io/s/pjj76x7o9q)

grsabreu:
any news here?

peterpme:
Hey @Hehk in an effort to clean up the repo, are you ok with closing this out and re-opening if its still valid?

Thanks!

Hehk:
@peterpme No problem.

peterpme:
Thanks! Closing this out for now :smile:

PR #231, opened 2018-05-27
✨ link to main reason site

alex-wilmer:
Haha was mildly annoyed by what feels like an obviously needed to link to the main reason site. Tried to make it an icon at first but got annoyed again and pushed this to see if maybe it's been there all along and I just don't see it.

cheers
a

🙃

### DIFF
--- a/website/i18n/en.json
+++ b/website/i18n/en.json
@@ -49,6 +49,7 @@
     "Community": "Community",
     "Blog": "Blog",
     "GitHub": "GitHub",
+    "Reason": "Reason",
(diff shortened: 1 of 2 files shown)


peterpme:
Hey @alex-wilmer,

Thanks so much for this! Would you be willing to update your PR and we can merge it in? Sorry for the delay 😅

rickyvetter:
I think that there are too many links in navigation for this to be added, but agree that more connection would be nice. I think on the main site having a button/link under "Safe and Sound" could be one approach. Open to others as well.

PR #238, opened 2018-06-05
Fix braces typo in docs Callback Handlers

mudrz:
Replace braces: `render: (self) => <button onClick={self.handle(click)} />` should be     `render: (self) => <button onClick=(self.handle(click)) />`


peterpme:
Hey @mudrz,

Thanks for this PR and so sorry about the delay 😅 

Since this API is deprecated I'm going to close this out. That being said, if you can update this with the newer api in mind, please re-open.

Thanks!

PR #241, opened 2018-06-09
algolia doc search for many languages

endiliey:
## Notes
Merge only after https://github.com/algolia/docsearch-configs/pull/447 is merged

## Motivation
Based on https://github.com/reasonml/reasonml.github.io/pull/392#issuecomment-395927699

We want to search the correct language in the search bar.

When I'm on chinese, I will only search for chinese documentation.
When I'm on english,, I will only search for user documentation.

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -217,7 +217,10 @@ const siteConfig = {
   },
   algolia: {
     apiKey: "55156da6520de795d3a2c2d23786f08e",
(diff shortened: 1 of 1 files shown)


s-pace:
Done from our side

chenglou:
Thanks =D

endiliey:
screenshot_20180612-174715

Should work. Let me know if there is any problem in the future.

PR #242, opened 2018-06-09
Rename ReasonReact module to React

jchavarri:
An idea attempting to improve the dev exp a little bit.

This is one of the most used bindings in the Reason ecosystem and the only one (afaict) that includes `Reason` in the main exported module name. This PR proposes removing the `Reason` prefix from the module name.

I left the docs out for now... until getting some feedback on the feasibility of it.

A `ReasonReact.re` module is left for backwards compat, which just includes the `React.re` one.


cullophid:
Im already doing this in every project

rrdelaney:
I think the React global namespace is already used by another project on Opam: https://opam.ocaml.org/packages/react/

However, I don’t think there’s any projects using ReasonReact and React (the OCaml one), so it’s probably not a big deal? I don’t know, just adding some context 🙂

jordwalke:
@rrdelaney How about `Reaccs`? 😆
FWIW, we're already using `React` internally at Facebook, too.

rrdelaney:
Clarifying my comment, I don't think the name clash would be a problem, and I'm very much in favor of this change 😄I just remembered that someone brought up the OPAM React once, and I figured I'd add the context that I knew

ulrikstrid:
Since #351 does this maybe this issue can be closed?

PR #244, opened 2018-06-11
Try ReasonReact at Repl.it

amasad:
Link here: https://repl.it/languages/reactre


rrdelaney:
This is amazing :heart: Is reason-scripts special-cased at all? I'll have to try out new releases on this first :smile:

amasad:
@rrdelaney you can simply update the package.json and just restart the server 👍

rickyvetter:
Hey - sorry for the delay on this. Would love to add but was trying to figure out how to change the default dependencies for https://repl.it/languages/reactre - is it possible to do on our end?

amasad:
Hey @rickyvetter / @peterpme! Since this PR we actually moved away from adding custom languages and towards a template-based approach. That's better because if you want to update the packages you can just do it from `package.json`, our package manager UI, or npm.  I just created a quick template for Reason React but I'd prefer that you fork it and own it so you can maintain it: https://repl.it/@amasad/ReasonReact-Template

When you're ready I can just redirect repl.it/l/reactre to this template like we do with the jest team: http://repl.it/languages/jest

peterpme:
Thanks @amasad !!

You can point it to this one: https://repl.it/@reasonreact/ReasonReact-Template

PR #245, opened 2018-06-12
Add click counter example; rename timer example

gilbert:
This fills in a small void. It's more complicated than the simple example, but simpler than [the reducer docs](docs/state-actions-reducer.md) without introducing timeout management in the current counter example.

I renamed the current counter example to `timer.md` to give it a more appropriate name ("counter" is more common for a click-handling component), and replaced `counter.md` with this new example to avoid any broken links.

### DIFF
--- a/docs/counter.md
+++ b/docs/counter.md
@@ -1,64 +1,36 @@
 ---
-title: Counter
+title: Click Counter
 ---
 
-Demonstrates calling an async "setState" in ReasonReact.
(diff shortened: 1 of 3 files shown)


rickyvetter:
Thanks so much for this contribution and so sorry for the lack of response. Unfortunately this PR now has conflicts with master and makes use of some of ReasonReacts older and dis-recommended APIs. Closing this out as it isn't actionable but happy to chat about adding this with the new Hooks api.

PR #251, opened 2018-06-24
Update Docusaurus to 1.3.0

endiliey:
## Summary
Docusaurus 1.3.0 brings about some [backwards-incompatible CSS](https://github.com/facebook/Docusaurus/blob/master/CHANGELOG.md#breaking-changes) changes. So the team is helping the major users update 😄

Other changes
- Activated `cleanUrl` / pretty url for reason-react. All url is now accessible without `.html` extension
- Remove link with `.html` to without `.html`
- Removed some custom CSS pertaining to the side nav as the fixed nav is now part of Docusaurus core. Also changed the code blocks in table markup and standardized them across the site.

cc @JoelMarcey @yangshun 

## Test Plan
Ran the site locally and visited every page.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -17,7 +17,7 @@
     "url": "git+https://github.com/reasonml/reason-react.git"
   },
   "dependencies": {
-    "docusaurus": "1.1.5"
+    "docusaurus": "1.3.0"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


chenglou:
Thanks a lot =D =D

PR #253, opened 2018-07-03
Update link in event.md to point to master version that uses the newer Reason syntax

joealden:
Hello,

Before this change, the link to the inline docs pointed to an older version of ReactEventRe.rei. This version of the file used the older Reason syntax. This change updates the link to point to the version of this file in master. This means that this issue will not happen as the link will always point to the latest version of the file that is in master.

### DIFF
--- a/docs/event.md
+++ b/docs/event.md
@@ -2,7 +2,7 @@
 title: Event
 ---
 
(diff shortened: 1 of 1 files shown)


chenglou:
Thanks! This was intentional, as we didn't want to have files change under us while the documentation goes stale. But in retrospect, it doesn't make much sense here.

Merged!

PR #254, opened 2018-07-03
Fix typo in element-type-is-invalid.md

joealden:
Remove the duplicate 'you' that was present at the beginning of this page.


chenglou:
Thanks!

PR #255, opened 2018-07-04
Change intro to state-actions-reducer.md

dekkofilms:
I found a _possible_ typo to the intro paragraph in the documentation in this file. Was it intended to repeat `If this word doesn't mean anything to you`? Let me know if this is a rather small fix, but just wanted some clarity around it. By the way, this is great! :)


cristianoc:
Thanks

PR #259, opened 2018-07-12
Update example to working component.

seethroughtrees:
The current documentation uses the [deprecated](https://reactjs.org/blog/2017/04/07/react-v15.5.0.html) `createClass` method, which throws an error.  
Rather than adding another dependency, it seems easier to use a stateless react component.  

credit to: https://reasonml.chat/t/typeerror-react-createclass-is-not-a-function/388/6


aij:
> Additionally, you can't return null from a component

Why do you say that?

chenglou:
Thanks!

PR #260, opened 2018-07-20
Use ReasonReact.string instead of stringToElement

yrezgui:
As `ReasonReact.stringToElement` is deprecated, we should update our examples to reflect the new API


chenglou:
Hah, thanks!

PR #267, opened 2018-08-13
Docs: Remove mention of "two solutions" for install

jbr:
As of 8c6b77a0d379244d8d0631a7089d40d1be1f0b30, there's only one solution suggested by the docs


chenglou:
Thanks!

PR #272, opened 2018-08-27
docs: change Update usage to Reason 3 syntax

zploskey:
This `ReasonReact.Update` example was overlooked in the update to Reason 3 syntax.


chenglou:
Thanks

PR #277, opened 2018-09-19
[RFC] Experiments with the API

bloodyowl:
I made a few explorations on how it'd be possible to fix some of the things in the FUTURE.md file

## stateless ≠ reducer

Stateless and reducer here have different types. This helps as it prevents settings non-existant fields to `statelessComponent`.

The main `component` type is the following:

```reason
type component('a) =
  | Stateless(statelessComponent): component(statelessComponent)
  | Reducer(reducerComponent('state, 'action)): component(reducerComponent('state, 'action));
```

A drawback of this is that you need to specify the type returned by the `make` function:

```reason
let component = ReasonReact.statelessComponent("Counter");
let make = _ =>
  ReasonReact.Stateless({
    ...component,
    render: _self => <div> "Helloworld"->ReasonReact.string </div>,
  });
```

I couldn't find a sound way to keep the `component` spread without adding this. If there's some, that'd be super interesting.

## remove retainedProps and willReceiveProps in favour or getDerivedStateForProps

```reason
type state = {
  last: option(int),
  value: int,
};
let component = ReasonReact.reducerComponent("Counter");
let make = (~value, _) =>
  ReasonReact.Reducer({
    ...component,
    initialState: () => {last: None, value},
    reducer: ((), _) => NoUpdate,
    getDerivedStateFromProps: state => {
      last: state.value !== value ? Some(state.value) : state.last,
      value,
    },
    render: ({state}) =>
      <div>
        (
          "last : "
          ++ state.last
              ->Belt.Option.map(string_of_int)
              ->Belt.Option.getWithDefault("Nothing")
        )
        ->ReasonReact.string
        <br />
        ("current : " ++ state.value->string_of_int)->ReasonReact.string
      </div>,
  });
```

## make unset lifecycle methods undefined on React side

this was a bit tricky: because we need to know what's in the component spec to set the prototype, the `initialState` checks if it's the first call (using a `__defined` in the component's prototype) and sets the methods. this is needed in order to get the spec value returned by the user's `make`, not the default one.
 
## add didCatch

```reason
type action =
  | HandleError(string);
type state = {error: option(string)};
let component = ReasonReact.reducerComponent("DidCatch");
let make = children =>
  ReasonReact.Reducer({
    ...component,
    initialState: () => {error: None},
    reducer: (action, _state) =>
      switch (action) {
      | HandleError(string) => Update({error: Some(string)})
      },
    didCatch: ({send}, exn, info) => {
      Js.log(info.componentStack);
      switch (exn) {
      | Invalid_argument(value) => send(HandleError(value))
      | _ => ()
      };
    },
    render: ({state}) =>
      <div>
        {
          switch (state.error) {
          | Some(error) =>
            ("An error occured: " ++ error)->ReasonReact.string
          | None =>
            switch (children) {
            | [|child|] => child
            | _ => ReasonReact.null
            }
          }
        }
      </div>,
  });

```

## simpler createClass

In order to set `getDerivedStateFromProps` as a static prop, I had to make a simple `createClass` to replace the optimizedCreateClass that was there.


chenglou:
cc @rickyvetter

peterpme:
Hey @bloodyowl,

Thanks so much for this PR and sorry about the delay! In an effort to clean the repo up (and b/c its using the old api) I'm going to close this for now.

Thanks!

PR #280, opened 2018-09-28
Introduce ReasonReact.Router.replace

### LINKED ISSUE #278: Router does not support `replace`
The browser history object exposes both `push` and `replace` capabilities for changing the url but the Router module only exposes the `push` function.

The [`replace`](https://developer.mozilla.org/en-US/docs/Web/API/Location/replace) function is desirable in a common application flow for authorization.

I want some of my routes to only be accessible by logged in users. Imagine a user is on the home page (`/`) and clicks a link to a protected page (`/protected`). If the user is not authorized, they should be redirected to a login page(`/login`). Once on `/login`, if the users presses the back button, they should be taken to `/`, not `/protected` (which would just redirect again to `/login`).

gavacho:
Adds `replaceState` where there was only `pushState`. Fixes #278

### DIFF
--- a/lib/js/src/ReasonReact.js
+++ b/lib/js/src/ReasonReact.js
@@ -490,6 +490,18 @@ function push(path) {
   }
 }
 
+function replace(path) {
+  var match = typeof (history) === "undefined" ? undefined : (history);
+  var match$1 = typeof (window) === "undefined" ? undefined : (window);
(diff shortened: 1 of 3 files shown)


peterpme:
:wave: anything I can do to help move this along?

chenglou:
Looks good. Thanks! Sorry for the late merge.

Issue #137, opened 2017-12-07
Inconsistency between min and max type


thangngoc89:
Currently, min accepts `int` and max accepts `string`. I think they should have the same type which is `string`.

This is a valid syntax: ` <input type="date" name="bday" min="2000-01-02">` 

@chenglou I can send a quick PR to fix the type of min to string if you want. Could you confirm on this?

chenglou:
Sure! Thanks.

dusterherz:
Any news on this issue ? Had the same problem with date inputs.

peterpme:
For the sake of cleaning up the repo (and given how old this issue is) I'm going to close this out.

Please re-open if this is still relevant. Thanks!

PR #281, opened 2018-09-30
set "min" property to "string"

stuartkeith:
Related to issue #137.

The `min` property is currently set to be an int - however `min` is used for a variety of input types (number, range, date, etc), and `int` is not always compatible.

This PR sets it to be `string` instead.


sgrove:
Looks like this might be superseded by #442 ?

PR #284, opened 2018-10-08
Update Router doc grammar

dangdennis:


dangdennis:
Changing text from `but to arrived at it` to `but to have arrived here,`

chenglou:
Thanks!

PR #290, opened 2018-10-23
crossorigin case and type incorrect

Enalmada:
Currently crossorigin but should be crossOrigin to avoid 
> Warning: Invalid DOM property `crossorigin`. Did you mean `crossOrigin`?

Should be string to avoid:
> If you want to write it to the DOM, pass a string instead: crossOrigin="true" or crossOrigin={value.toString()}.


chenglou:
Thanks!

PR #297, opened 2018-11-02
Add integrity hashes in yarn.lock

MoOx:
Recent version of yarn (1.10+?) add integrity hashes in the lock file. I guess we should have those.


peterpme:
Hey @MoOx,

Thanks so much for this change! Do you mind re-opening with the update?

peterpme:
Update: since we've got the latest version of yarn running, going to close this out for now. Thanks!

PR #298, opened 2018-11-03
Uncurry jsPropsToReason

jenriquejr:
- avoid curry overhead when calling jsPropsToReason
- fixes issue with react-hot-loader

### DIFF
--- a/lib/js/src/ReasonReact.js
+++ b/lib/js/src/ReasonReact.js
@@ -360,8 +360,9 @@ function element($staropt$star, $staropt$star$1, component) {
 }
 
 function wrapReasonForJs(component, jsPropsToReason) {
+  var uncurriedJsPropsToReason = Curry.__1(jsPropsToReason);
   var tmp = component[/* reactClassInternal */1].prototype;
-  tmp.jsPropsToReason = Js_primitive.some(jsPropsToReason);
+  tmp.jsPropsToReason = Js_primitive.some(uncurriedJsPropsToReason);
   return component[/* reactClassInternal */1];
 }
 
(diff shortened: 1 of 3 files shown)


jenriquejr:
Tools like react-hot-loader wrap react class methods on proxies, therefore if jsPropsToReason is curried it doesn't execute and ultimately an error is thrown.

This commit fixes the problem.

screen shot 2018-11-03 at 6 05 08 pm

cristianoc:
The fix is legit.
The downside is that all end users need to learn, and use, uncurried calls.

+ @chenglou @rickyvetter

chenglou:
That's unfortunate... it's a needed change, but we can't do it right now in the short-term, since that's a breaking change that affects quite a few components, especially when folks drag in third-party. Will think about it, thanks.

jenriquejr:
agreed @cristianoc - unsure about how I could make this type not be public
I attempted using the `pri` keyword but the result is not what I was looking for in this case.
the .rei file must match the api that is being exported so I had to include it there.
I'll see if I can get it to work with a few more attempts

suggestions appreciated -

cristianoc:
@jenriquejr I mean literally remove this line from the .rei file:
```reason
type uncurriedJsPropsToReason('jsProps, 'state, 'retainedProps, 'action) =
  (. 'jsProps) => component('state, 'retainedProps, 'action);
```
Does it compile?

jenriquejr:
it does @cristianoc !!! at first I thought it did not !! thanks

cristianoc:
Thank you so much for the awesome fix. Merging now.

PR #305, opened 2018-12-05
Specify which index.html to open

prestontw:
In the Installation page, it says that that you can open `index.html`, but doesn't specify where this is. People not familiar with React/webpack might think that this is referencing `src/index.html`, which is a small but confusing deterrent from getting a new project in a new technology started.


rickyvetter:
Added in https://github.com/reasonml/reason-react/commit/7062d0cd97bda273c3c928496c83d729a5457495 - thanks for the callout!

PR #311, opened 2018-12-24
Add Literal to projects

javamonn:
[Literal](https://literal.io/) uses `reason-react` for all new development as well as several existing projects, including the PDF reader itself.

We've open sourced several projects and bindings:
- https://github.com/literal-io/calvino
- https://github.com/literal-io/bs-pdfjs
- https://github.com/literal-io/bs-history

 Thank you for your work on `reason-react`!


chenglou:
Welcome!

PR #313, opened 2019-01-09
Clarify wording in callback-handlers.md

jey:
Should be phrased as "reading from" here, though "writing into" makes sense for cases of mutation.


chenglou:
Thanks!

PR #316, opened 2019-01-21
Clarify initial app entry point in docs

lukeknoot:
This may not be the exact intention of the wording, depending on whether it was specific to the output of bsb, or was indeed about the app's entry point. 

However the combination of reading the homepage stating `src/` and the installation page stating `index.html` lead me to think the index.html was directly openable from `src/`, when it's really `build/`.

I don't mind if this gets addressed in #305.


rickyvetter:
Looks like it was added in https://github.com/reasonml/reason-react/commit/7062d0cd97bda273c3c928496c83d729a5457495

PR #331, opened 2019-02-12
Use bs.get/bs.set for ref.

cknitt:
@rickyvetter: As discussed on Discord.


chenglou:
Thanks!
@rickyvetter what's this?

PR #336, opened 2019-02-16
Add React.Context.Provider.

cknitt:
@rickyvetter Here is a Context.Provider binding as discussed on Discord.

Should we also change

* `React.context` -> `React.Context.t`
* `React.createContext` -> `React.Context.make`

?

### DIFF
--- a/lib/js/src/React.js
+++ b/lib/js/src/React.js
@@ -3,5 +3,13 @@
 
 var Ref = /* module */[];
 
+function Provider(C) {
+  var make = C[/* context */0].Provider;
+  return /* module */[/* make */make];
+}
+
+var Context = /* module */[/* Provider */Provider];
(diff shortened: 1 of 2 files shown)


lib/js/src/React.js
```diff
@@ -3,5 +3,13 @@
 
 var Ref = /* module */[];
 
+function Provider(C) {
```
rickyvetter: There has got to be a way to eliminate this runtime. I really would love to avoid this.

cknitt: I would love that, too! But I am not aware of any way around this, unfortunately. :-(

rickyvetter:
kk. Not committing to this long term because I think a cleaner solution might present itself, but I think this works for now.

PR #338, opened 2019-02-17
useUrl hook for router

cknitt:


src/ReasonReact.re
```diff
@@ -883,6 +883,20 @@ module Router = {
...
+
+  let useUrl = () => {
+    let (url, setUrl) = React.useState(dangerouslyGetInitialUrl);
+
+    React.useLayoutEffect1(
+      () => {
+        let watcherId = watchUrl(url => setUrl(_ => url));
+        Some(() => unwatchUrl(watcherId));
+      },
+      [||],
```
rickyvetter: `use(Layout)Effect0` should get you this effect already.

src/ReasonReact.rei
```diff
@@ -254,6 +254,8 @@ module Router: {
       for an example.
       */
   let dangerouslyGetInitialUrl: unit => url;
+  /** hook for watching url changes */
+  let useUrl: unit => url;
```
rickyvetter: Offline you were mentioning that you think we should move this into a separate module to get the dependencies correct. How exactly were you thinking about doing that? Want to make sure we have a plan for this before we add more stuff here.

src/ReasonReact.re
cknitt: Ah, right! That's convenient!

src/ReasonReact.rei
cknitt: IMO the `ReasonReact.re` as it is now on the hooks branch should be split into three parts:

1. "Classic" ReasonReact (record-based) components
2. The router
3. The `useRecordApi` compatibility hook

If you don't want a breaking change, we could move the contents of the `Router` submodule into `ReasonReactRouter.re` and have `module Router = ReasonReactRouter;` in `ReasonReact.re`.

`useRecordApi` could be moved into `ReasonReactRecordApi.re` or `ReasonReactCompat.re` or something like that.

This way new projects using the hooks API only can just use `React.re` and, if they need it, `ReasonReactRouter.re` without pulling in "legacy" stuff.

But this can be done in a separate PR after `useUrl` is added.

rickyvetter: Got it. Yeah I like `ReasonReactCompat.re` and `ReasonReactRouter.re` with a module alias for backwards compat. Kind of a shame about the collision with ReactRouter. Let’s go with this and talk to Cheng in the next PR about where he wants that one.

src/ReasonReact.re
```diff
@@ -883,6 +883,17 @@ module Router = {
     | Some((window: Dom.window)) =>
       removeEventListener(window, "popstate", watcherID)
     };
+
+  let useUrl = () => {
+    let (url, setUrl) = React.useState(dangerouslyGetInitialUrl);
+
+    React.useLayoutEffect0(() => {
```
rickyvetter: Why did you opt for LayoutEffect here? I think that because this doesn’t need to be synchronous Effect might be a better choice, but I’m not sold either way really. I don’t think this is blocking this PR but since it would technically be a breaking change I think we should resolve before a wider release.

cknitt: The use case I had was only working with `useLayoutEffect`.

(In my app, for some routes, I would render a `Redirect` component that would push a new route. For some reason, this only worked when using `useLayoutEffect` here and `useEffect` in the `Redirect` component. I did not investigate it any further.)

rickyvetter: Interesting. LayoutEffect happens earlier than Effect so perhaps these hooks are called upward and this order is required in order to trigger this subscription before the Redirect component executes. That seems like a reasonable excuse to put this here.

PR #340, opened 2019-02-20
Update Test.re

arbipher:
replace React with ReasonReact.
Since ReasonReact is defined as an alias for React, I think we should always use this as module name.


peterpme:
Hey @arbipher,

Thank you so much for this PR and sorry about the delay 😅

In an effort to clean this up I'm going to close this out. If you'd like to re-open it updated against master, please feel free.

Thank you!

PR #343, opened 2019-03-06
Fix type of Context.provider.

cknitt:


rickyvetter:
Thanks! This is much more accurate.

PR #345, opened 2019-03-07
Tests for ppx

jchavarri:
I was looking at the new components ppx and I wasn't sure what exactly it was doing, so I tried to set up some tests with Dune to see the output more clearly, and I thought this might be useful for others.

There is an input file `test.ml` (I haven't managed to make it work with `.re` files) and the expected output in `pp.expected`.

The ppx itself is mostly untouched, just separated into two files: one with the lib (`jsx_ppx/lib/reactjs_jsx_ppx_v3_lib.ml`) and another where the driver and standalone are created (`jsx_ppx/bin/reactjs_jsx_ppx_v3.ml`).

The diff runs with:

```
$ esy test
```

Happy to take it through the finish line if this is useful, I'd need some guidance with the `.re` compat though :) I guess `refmt` could be used to preprocess the test file?


jsx_ppx/test/test.ml
```diff
@@ -0,0 +1,12 @@
+[@@@bs.config { jsx = 3 }]
+
+let div ~children  () = children
+
+module Test = struct
+
+let make ~a  ~b  _ =
+  print_endline (a ^ b);
+  print_endline "This function should be named `Test`";
+  ((div ~children:[] ())[@JSX ])[@@react.component ]
```
jchavarri: Besides this file ideally being using Reason syntax, I wonder if using JSX format (`<div />`) would be possible too 🤔

rickyvetter:
Nice! Thank you. - couple questions (I'm not super familiar with dune):

Why setup as a lib? Is this better practice than just testing the executable? It seems like that is what makes it tricky to support Reason syntax in the tests.

How would I set this up if I wanted like 20 different test files? Do I need to update the dune file to reflect each of those?

The main thing I hate are these .opam files cluttering everything up. 

Unrelated to dune - where do you want the canonical home of this ppx to be?

jchavarri:
> Why setup as a lib? Is this better practice than just testing the executable? It seems like that is what makes it tricky to support Reason syntax in the tests.

Some context is needed: I am trying to migrate reason-react [to jsoo](https://github.com/jchavarri/rroo/). I started with v2, but it was too complex and I gave up. But now that the bindings got much simpler with hooks I'm trying again (that `rroo` repo has not been updated yet as I'm just exploring locally).

The main reason why I thought a library would be a good idea is because I believe that for compiling with dune (what jsoo uses) it is required that ppxs are packaged as libraries that use omp drivers: https://jbuilder.readthedocs.io/en/latest/dune-files.html#preprocessing-with-ppx-rewriters.

So, having the ppx as a library would allow for this integration with dune + keep producing the executable for BuckleScript. But I see how it might be too much overhead at this point.

> How would I set this up if I wanted like 20 different test files? Do I need to update the dune file to reflect each of those?

I have no idea. In theory dune's usage of `diff` is using just the underlying `diff` which should support folders, but idk how to avoid creating a rule for each file. I asked on the OCaml Discord for help.

> The main thing I hate are these .opam files cluttering everything up.

You mean the ones in `esy.lock` folder? Those should be automatically handled by esy. Or are you referring to the extra `.opam` files needed for the lib + executable? If the library was removed (until the jsoo story is more clear) and there was only the executable, then no extra opam files would be needed.

> where do you want the canonical home of this ppx to be?

I'm not sure if you're referring to which repo? I think this repo is the most natural place to have it because this ppx doesn't make sense outside ReasonReact bindings. Hopefully one day BuckleScript gets support for the more traditional dune + esy workflow, so adding the ppx in a project would be just a couple of extra lines in `package.json` and `dune` files, and you guys can avoid having to copy the ppx over for each release (which is much appreciated now! 😄 ).

PR #346, opened 2019-03-08
Make Fragment children non-optional.

cknitt:
IMHO a Fragment without children does not really make sense.


rickyvetter:
Closed with 1cb0fde0a1f2110d5f4f6e70d9eb04abd7679141

PR #349, opened 2019-03-11
Add currying friendly hooks

bordoley:
I've used these wrappers for useMemo and useEffect fairly extensively and find that they greatly improve the code quality of my components by allowing me to lift callback functions out of the render function. Hence I wanted to propose them for inclusion in reason-react proper. A common way of writing components that use these callback is as follows:

```
[@react.component]
let make = {
  let createOnClick = (prop1, prop2, ev) => {
    // use the event and props to do perform some side effect
    ()
  };

  (~prop1, ~prop2) => {
    let onClick = ReactHooks.useMemo2(createOnClick, prop1, prop2);
    <button onClick>{React.string("hello world")}</button>;
};


chenglou:
Hey Dave. We're gonna hold this off. Hooks not released yet

chenglou:
Humm this is a bit too stale, so I'll close it, sorry. Additionally we explicitly don't wanna make the hooks currying-friendly. It has a non-trivial perf cost, plus code understanding cost. Not to mention whatever future react hooks semantics that might not work with currying.

PR #355, opened 2019-03-15
Attempt an extremely hacky solution for setting a default component instead of using "make"

phated:
I had some time today so I wanted to see if I could expand on [my comment](https://github.com/reasonml/reason-react/pull/351#issuecomment-473060066) about flagging a function as the "default" so the `make` name could be available for other things within a module.

I have no delusions that this code will land, but I thought it would be a good thought exercise to see if we could alter the PPX in some way support this feature.

It took me **waaaaay** too long to understand the the `[@react.component]` attribute and the JSX transform were being handled completely independently.  Once that clicked, I figured we'd need some sort of PPX-global state that would store the re-mapped default identifiers. 

Based on the tests I added, this seems to be working but I have no idea what it would do to performance or if this is a massive anti-pattern for PPXs.

cc @rickyvetter

### DIFF
--- a/jsx_ppx/reactjs_jsx_ppx_v3.ml
+++ b/jsx_ppx/reactjs_jsx_ppx_v3.ml
@@ -78,6 +78,7 @@ type 'a children = | ListLiteral of 'a | Exact of 'a
 type componentConfig = {
   propsName: string;
   forwardRef: string option;
+  isDefault: bool;
 }
 
 (* if children is a list, convert it to an array while mapping each element. If not, just map over it, as usual *)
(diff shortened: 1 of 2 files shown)


rickyvetter:
So this code is run in a different process per file. This works for a single file compilation but would actually not be sufficient for a project. You would need to persist this map to a file and manage deletions and all the fun stuff that comes with build systems. I don’t think this complexity is worthwhile.

phated:
I knew I was missing something fundamental 😂

rickyvetter:
We talked about this offline and I think even if we do want to allow this kind of configuration we'd need to go a different route.

PR #356, opened 2019-03-19
Begin refactoring & annotating the PPX

phated:
@rickyvetter I wasn't able to get as far as I would have liked on this today but I figured I'd send some of the stuff I saw while working on my last PR as a "draft PR".   Please let me know if this looks good and I'll continue with updates tomorrow.

Also, would it make sense to add `ocamlformat` to the project?


rickyvetter:
I'm taking this even though you haven't said it's good to go yet! I think it's definitely an improvement and I need to put some more stuff on top of it.

PR #358, opened 2019-03-26
add possiblity to set a static url

sync:
@rickyvetter  what do you think of this ?


bloodyowl:
what would you think of naming the parameter `initialUrl`?

sync:
discussion is in the hooks branch, also was added directly with this commit https://github.com/reasonml/reason-react/pull/351/commits/bfc9386dd49f1639c42806f623c903ea602bdb6b

PR #361, opened 2019-03-29
Move react and react-dom into peerDependencies

sync:
Avoiding ending up with multiple versions of react in one repository when using reason-react.

For example if you use the latest master version react and you are using latest react in your repo:

```
"dependencies": {
    "react": "16.8.6",
    "react-dom": "16.8.6",
    "reason-react": "git://github.com/reasonml/reason-react.git"
  },
```

You'll end up with 2 different versions of React.

You'll then encounter this book in your code if you are using hooks:

```
 Invariant Violation: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following
 reasons:
    1. You might have mismatching versions of React and the renderer (such as React DOM)
    2. You might be breaking the Rules of Hooks
    3. You might have more than one copy of React in the same app
    See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.

      29 |           /* hash */"",
      30 |           /* search */""
    > 31 |         ], /* () */0) : ReasonReactRouter.useUrl(undefined, /* () */0);
         |                                           ^
      32 |   var match = mapPathToRoute(url[/* path */0]);
      33 |   return React.createElement("div", undefined, React.createElement(GraphqlHooks.ClientContext.Provider, {
      34 |                   value: client,
```

If you are a yarn user you can solve it by doing this:

```
"resolutions": {
    "react": "16.8.6",
    "react-dom": "16.8.6"
  },
```


sync:
@rickyvetter what do you think ?

RomanGalochkin:
Please, merge this PR.

tmattio:
I think this will solve #447 as well.

tizoc:
Just hit this too.

tizoc:
Just switched to Yarn as suggested [here](https://reactjs.org/warnings/invalid-hook-call-warning.html#duplicate-react). As a workaround, it works.

peterpme:
Hey @sync 

Thanks so much for this! Do you mind updating this against master and we'll be able to merge in?

Sorry about the delay 😅

sync:
will do this asasp

sync:
@peterpme done thanks for taking care of this

sync:
if i can help doing anything else let me know

PR #363, opened 2019-04-03
Expose ariaChecked and ariaInvalid

bsansouci:
Not sure if there's a reason to expose a more strict API tbh.
The alternative is exposing: 
```reason
`grammar | `spelling | `True | `False
```
which is kinda meh too


rickyvetter:
I'm still holding out for a better solution here but recognize this is a common and very real issue. If we don't have something better by 1.0 this will go out.

davesnx:
Closing in favor of https://github.com/reasonml/reason-react/pull/697

PR #375, opened 2019-04-11
Update homepage BSB theme

MarcelCutts:
Real quick - this updates the instructions on the homepage to use the new hook-focused React theme. 

As the first example and docs primarily focus on this theme now, it seems correct for this to be consistent as a first impression to the users.


rickyvetter:
Good call!

PR #376, opened 2019-04-11
Add createRef

bloodyowl:
Can be useful when dealing with the old API. 

e.g.

```reason
type state = {
  containerRef: ReactDOMRe.Ref.currentDomRef
};

let state = {
  containerRef: React.createRef()
};

<div ref={state.containerRef->ReactDOMRe.Ref.domRef} />
```

### DIFF
--- a/src/React.re
+++ b/src/React.re
@@ -25,6 +25,8 @@ module Ref = {
   [@bs.set] external setCurrent: (t('value), 'value) => unit = "current";
 };
 
+[@bs.module "react"] external createRef: unit => Ref.t('a) = "";
+
 module Context = {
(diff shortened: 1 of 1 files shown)


src/React.re
```diff
@@ -25,6 +25,8 @@ module Ref = {
   [@bs.set] external setCurrent: (t('value), 'value) => unit = "current";
 };
 
+[@bs.module "react"] external createRef: unit => Ref.t('a) = "";
```
rickyvetter: This needs to be `unit => Ref.t(Js.nullable('a))` or else I don't think it's safe.

PR #380, opened 2019-04-12
Make props-spread docs friendlier

jacobp100:
I think there are some legitimate use-cases for spread props, so I wanted to rephrase the issue in a more positive way.

I'm also not sure the link is still relevant with the new API. But I wasn't sure, so kept it in anyway

### DIFF
--- a/docs/props-spread.md
+++ b/docs/props-spread.md
@@ -2,4 +2,10 @@
 title: Props Spread
 ---
 
-You can't. Props spreading is a big source of unpredictability and performance regression (think `shouldComponentUpdate`). Our API prevents this. If you reaaaaally need it for binding to existing ReactJS components, see [this section](clone-element.md).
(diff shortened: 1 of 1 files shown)


justgage:
Sorry, just a random person here, no power to get this merged, but I think this is a huge improvement.

docs/props-spread.md
```diff
@@ -2,4 +2,10 @@
 title: Props Spread
 ---
 
-You can't. Props spreading is a big source of unpredictability and performance regression (think `shouldComponentUpdate`). Our API prevents this. If you reaaaaally need it for binding to existing ReactJS components, see [this section](clone-element.md).
+There is currently no story around doing this in a type-safe manner.
```
phated: Maybe something like "There is currently no way to do this in a type-safe manner." - I'm not sure the "no story" phrase translates well.

PR #391, opened 2019-04-15
Add docs and strategies for migration to v2 to v3 of JSX

jchavarri:
My attempt to document v2 to v3 migration approaches.

cc @rickyvetter @yawaramin @alexfedoseev in case you have any ideas to improve it, thanks!

### DIFF
--- a/docs/jsx.md
+++ b/docs/jsx.md
@@ -188,6 +188,7 @@ let make = (~src) => <img src />;
 
 module Jsx2 = {
   let component = ReasonReact.statelessComponent("Image");
+  /* `children` is not labelled, as it is a regular parameter in version 2 of JSX */
   let make = (~src, children) =>
     ReasonReactCompat.wrapReactForReasonReact(
       make,
--- a/docs/jsx.md
+++ b/docs/jsx.md
@@ -6,6 +6,8 @@ Reason comes with the [JSX](https://reasonml.github.io/docs/en/jsx.html) syntax!
 
 [Version 2](jsx-2.md) is used to render Reducer style components which require special interop to handle the ways they differ from ReactJS components. If you are starting a new project you can stick to version 3 all the time.
 
+For help to migrate from version 2 to version 3, check the section [below](https://reasonml.github.io/reason-react/docs/en/jsx.html#migrating-from-version-2-to-version-3).
+
 ## Uncapitalized
 
 ```reason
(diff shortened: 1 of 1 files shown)


docs/jsx.md
```diff
@@ -134,4 +136,81 @@ You're effectively passing the array `[| <div />, <div /> |]` to `MyReasonCompon
...
+```reason
+/* In Image.re */
+[@bs.config {jsx: 3}];
+
+[@react.component]
+let make = (~src) => <img src />;
+
+module Jsx2 = {
+  let component = ReasonReact.statelessComponent("Image");
+  let make = (~src, children) =>
```
jchavarri: @rickyvetter I'm not sure if `key` has to be added here?

rickyvetter: Nope. Not needed!

docs/jsx.md
```diff
@@ -134,4 +136,81 @@ You're effectively passing the array `[| <div />, <div /> |]` to `MyReasonCompon
...
+
+### File level
+
+For larger applications, it might not be possible to migrate all components at once. In these cases, a per-file migration is also possible.
+
+A file can be configured to use version 3 of the transform by adding `[@bs.config {jsx: 3}]` at the top of the file.
+
+The per-file configuration allows to mix, in the same application, components compatible with either of both versions of the JSX transforms. However, the restriction is that all the components used in a file will have to be compatible with the JSX version specified for that file.
+
+For example, if a file contains the following code:
```
rickyvetter: Yeah love having concrete examples here.

docs/jsx.md
diligiant: @jchavarri @rickyvetter rightfully insisted (Discord) that we added ~ to children.

alexfedoseev: @diligiant In this case—nope, since it's a wrapper to JSX@2 API: it requires unlabeled last argument.

diligiant: @alexfedoseev ok ; don't you think we need a section about these precise "rules" ? Once you know, you know but we might end up with troubles for newbies unless the rls emits a warning/error.

jchavarri: @diligiant I added an inline comment in https://github.com/reasonml/reason-react/pull/391/commits/36a4c98858c2077d9421bb22cea3913ae40efbe8 that explains why the label is not needed here. Please let me know if you have any specific suggestions on how to make it better.

diligiant: @jchavarri Thank you. It goes back to the difference with React where children is a special prop (3rd param of createElement) ; @rickyvetter was kind enough to highlight this to me.

Likewise, I don't think I've seen the explanation that `=?` inside an interop external is like v2 `[@bs.optional]`, different from normal `=?` (turns the param into `option(a')`).

docs/jsx.md
```diff
@@ -134,4 +136,82 @@ You're effectively passing the array `[| <div />, <div /> |]` to `MyReasonCompon
...
+
+```reason
+/* User.re */
+[@bs.config {jsx: 3}]
```
baransu: I think `;` is missing at the end of the line here 🙂

jchavarri: @baransu Oh, you're right, thanks! That's why I was getting warnings from the compiler 😅 fixed!

nikgraf:
@jchavarri sooo awesome 👍

PR #393, opened 2019-04-17
Add React.Children & cloneElement

bloodyowl:


src/React.re
```diff
@@ -27,6 +30,19 @@ module Ref = {
 
 [@bs.module "react"] external createRef: unit => Ref.t(Js.nullable('a)) = "";
 
+module Children = {
+  [@bs.module "react"] [@bs.scope "Children"] [@bs.val]
+  external map: (element, element => element) => array(element) = "map";
```
rickyvetter: Why does this return array instead of element? If there’s only one child does it wrap in an array?

bloodyowl: yeah, it does. there's a special case with `null` and `undefined` though, might be better to treat the return value as `element`.

PR #399, opened 2019-04-22
Remove maxDuration prop from suspense

anmonteiro:
As of https://github.com/facebook/react/pull/15272, `maxDuration` is no longer in React


rickyvetter:
Ugh kk. This is a breaking change. Will safe for the next release.

PR #400, opened 2019-04-22
Add Unstable module that can render in concurrent mode

anmonteiro:


src/ReactDOMRe.re
```diff
@@ -41,6 +41,41 @@ let renderToElementWithId = (reactElement, id) =>
   | Some(element) => render(reactElement, element)
   };
 
+module Unstable = {
+  type root;
+
+  [@bs.module "react-dom"]
+  external createRoot: Dom.element => root = "unstable_createRoot";
```
peterpme, line 48:
```suggestion
  external createRoot: Dom.element => root = "createRoot";
```

src/ReactDOMRe.re
```diff
@@ -41,6 +41,41 @@ let renderToElementWithId = (reactElement, id) =>
...
+  let createRootWithId = id =>
+    switch (_getElementById(id)) {
+    | None =>
+      raise(
+        Invalid_argument(
+          "ReactDOMRe.Unstable.createRootWithId: no element of id "
+          ++ id
+          ++ " found in the HTML.",
+        ),
+      )
```
rickyvetter, line 74: We just swapped to logging here instead of raising to avoid the additional dependency - could you move to that pattern here as well?

src/ReactDOMRe.re
rickyvetter: Just checked and it looks like they've removed unstable_ - but we should make sure our React versions are lining up for this change. https://reactjs.org/docs/concurrent-mode-reference.html#createroot

src/ReactDOMRe.re
```diff
@@ -41,6 +41,41 @@ let renderToElementWithId = (reactElement, id) =>
   | Some(element) => render(reactElement, element)
   };
 
+module Unstable = {
```
rickyvetter, line 44: Can we change the name to `Experimental` to match the React naming convention as well?

src/ReactDOMRe.re
anmonteiro: The other functions return `unit`. What do you suggest we return in the `None` case for this branch, given that the `Some` branch is returning a `root`?

anmonteiro: My suggestion is that the `createRootWith*` family of functions return `result(root, string)` type. That way we don't need to log, and can return something that the application can handle.

anmonteiro:
@rickyvetter @peterpme just rebased this and addressed your review comments.

I implemented the version I proposed in https://github.com/reasonml/reason-react/pull/400#discussion_r416180279 instead of throwing.

PR #401, opened 2019-04-22
Add missing use*N externals

alexfedoseev:
Let me know if you want to add `use*Unsafe` externals in this PR.

Also, I bumped bs-platform b/c after `yarn install` yarn introduced diff.


rickyvetter:
Thanks! Looks great.

PR #404, opened 2019-04-23
Add note on children in jsx migration section

jchavarri:
I figure this out after some time today and thought it could be useful for others.

@rickyvetter Does this approach make sense? Do you think we should include other cases we've discussed? (like `ref` and how it has different types in v2 and v3)

### DIFF
--- a/docs/jsx.md
+++ b/docs/jsx.md
@@ -213,6 +213,43 @@ let make = _children => {
 
 Once all components are using version 3, there is no more need to keep the `[@bs.config {jsx: 3}];` annotations at the top of each file, and they can be replaced by bumping the JSX version in the `bsconfig.json` file to `{"reason": {"react-jsx": 3}`.
 
+### Children
+
+Some components pass down `children` transparently, like:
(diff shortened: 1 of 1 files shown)


phated:
I like this but maybe it belongs in the documentation for `ReasonReactCompat` (which still needs to be written as per #372)

jchavarri:
@phated thanks for the heads up. I wonder if the section on "migrating from jsx versions" should be included there too?

https://reasonml.github.io/reason-react/docs/en/jsx#migrating-from-version-2-to-version-3

It seems there is quite a lot of information around the migration that it deserves its own section, as the issue you posted suggests.

rickyvetter:
@jchavarri any interest in making this new section? You can add in sidebars.json and then just make a new .md file. Also happy to merge this and then move it over!

jchavarri:
@rickyvetter I've proposed the new section for `ReasonReactCompat` in #424 and moved the content from this PR over there, so closing! 🙂

Issue #402, opened 2019-04-22
Maybe let's make a better `useState`?


jaredly:
I've got this at the top of my files, because `useState` is so cumbersome in the "fully type safe" case. Would it make sense to upstream? I know y'all are excited about having `React.re` have no runtime cost, so maybe it could go in `ReasonReact`, and we can `[@@deprecated]` the `React.useState` to avoid confusion?
```
let useState = (initial) => {
  React.useReducer((_ignored, newState) => newState, initial);
};
```

justgage:
I kind of wish there was a `useState` that works like the normal one in React, then a `useStateFn`that does what the current one does. It would be easy to write an upgrade script to help with that. Naturally it would be easy to write the `useState` in terms of the `useStateFn` (the one that currently exists).

rickyvetter:
I think there's definitely a place for a function like this. Going through backlog of other issues/PRs right now but thinking about the right place/way to do this.

PS - so sorry for the horribly slow response here.

peterpme:
Hey @jaredly 

Based on this convo, do you mind opening a PR?

For the sake of cleaning up the repo (and given how old this issue is) I'm going to close this out.

Thank you and sorry about the delay. Please re-open if its still relevant!

erlandsona:
Sorry to necrobump this but ran into this today helping out my coworkers try and modify some existing ReasonReact code to use local state instead of reductive.

I think that `useState` has to make a choice about which JS signature to support in the external is generally fine. Where we got burnt was that the compiler is unifying the [`Js.t('a)`](https://github.com/reasonml/reason-react/blob/master/src/ReactEvent.re#L17) with the signature of the `setProperty` function returned by `useState` when it should be throwing a compile error.

I see in the documentation there's a comment about returning a `Dom.eventTarget`, curious when that update will land because that'd prevent this issue from happening in the future.

Regarding the Synthetic event issue which in #476 because of unification there's a number of places in code where we've used the setter like `e => e->ReactEvent.Form.target##value->setProperty` which incorrectly unifies the `'a` type parameter which is ultimately just a raw `string` with the `string => string` function the `setProperty` function expects but it works because React ultimately handles both cases.

PR #426, opened 2019-05-12
Add more useState flavours

jchavarri:
This is mostly a discussion / question PR, related to #402.

Something like `useStateValue` below is really useful when one starts prototyping a component, and it feels like ReasonReact should provide an out-of-the-box solution for it.

Maybe there is some typing edge case that I'm missing? I read [this comment](https://github.com/jchavarri/reason-react/blob/6ecf6ee6fdb4412114de9eacba1b3adc517eb387/src/React.re#L116-L121) but I can't figure out in which cases this would not be safe.

### DIFF
--- a/src/React.re
+++ b/src/React.re
@@ -123,6 +123,11 @@ module Suspense = {
 external useState:
   ([@bs.uncurry] (unit => 'state)) => ('state, ('state => 'state) => unit) =
   "";
+[@bs.module "react"]
+external useStateValue: 'state => ('state, 'state => unit) = "useState";
+[@bs.module "react"]
+external useStatePrevious: 'state => ('state, ('state => 'state) => unit) =
+  "useState";
 
 [@bs.module "react"]
 external useReducer:


rickyvetter:
This isn't typesafe if `'state` is a function. I don't think we want to support this API (nor do I think it's the right api in JS).

jchavarri:
@rickyvetter Could you help me understand how safety is broken, please? I'm having a hard time finding an example of how that'd look like, I thought that separating the different cases with one external for each would solve any typing issues. thanks!

jchavarri:
n/m, I discussed with @cristianoc and understood the issue is that if state is a function of type `unit => 'a`, then `useState` will "interpret" that is the "lazy" version of `useState` and it'll behave differently.

Alternatives:
1. Define `useStateValue` and `useStatePrevious` on top of `useState`, so the lazy version is always used, and the behavior on JS side remains always the same. I guess this is a no-go as there's no way (that I know of) to do this without giving up on the "zero-cost" bindings policy.
2. Keep the existing API.
3. Polish this PR and adopt something like this.

I (unsurprisingly 😄 ) think we should go with 3. This is an issue in ReactJS API design, and I don't think it's pragmatic that ReasonReact users pay the cost to avoid the confusion for such a niche case by being offered a subset of the original API, which also happens to be the one that is least commonly needed.

If anyone _really_ needs to store a function `unit=>'a` as state, they can use the already existing `useState`, but for all other cases, which will be the large majority, `useStateValue` and `useStatePrevious` are available (we can change/ improve the naming if needs be).

rickyvetter:
I can't justify shipping code that's unsafe in this way for people even in edge cases, so 3 really isn't an option for me. 1 is something that I am open to. A couple options would be to have another module like `React.Utils` or `ReactUtils` which builds off of the no runtime core implementation. I am uneasy about shipping with `reason-react` - would be interested in thoughts here.

Sorry for going silent on this and many other issues. Getting more involved here again.

PR #427, opened 2019-05-13
Uncurry useEffect3-7

### LINKED ISSUE #425: `useEffect3`–`useEffect7` bindings should use `[@bs.uncurry]`?
The bindings upto `useEffect2` take an auto-uncurried callback type as the first parameter, however the ones after that do not. Was this intentional, and if so why? If not, I suggest making them all uniformly auto-uncurry with `[@bs.uncurry]`. The function type itself can be elegantly lifted out because the annotation actually takes an integer payload indicating the function's arity, e.g.

```reason
// Types for descriptive purposes
type cancelEffect = unit => unit;
type effect = unit => option(cancelEffect);

[@bs.module "react"] external useEffect2: ([@bs.uncurry 1] effect, ('a, 'b)) => unit = "useEffect";
[@bs.module "react"] external useEffect3: ([@bs.uncurry 1] effect, ('a, 'b, 'c)) => unit = "useEffect";
// and so on
```

I'm not sure if this should be considered a breaking change or not, though.

alexfedoseev:
Somehow I missed that in the previous PR.
Fixes #425

### DIFF
--- a/src/React.re
+++ b/src/React.re
@@ -157,22 +157,30 @@ external useEffect2:
   ([@bs.uncurry] (unit => option(unit => unit)), ('a, 'b)) => unit =
   "useEffect";
 [@bs.module "react"]
-external useEffect3: (unit => option(unit => unit), ('a, 'b, 'c)) => unit =
+external useEffect3:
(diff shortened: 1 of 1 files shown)


rickyvetter:
I don't think lifting this will work with bs.uncurry.

PR #433, opened 2019-05-24
Add binding for name and displayName

czystyl:
I added binding for `name` and `displayName` of the React Component.


src/React.re
```diff
@@ -390,3 +390,10 @@ external useImperativeHandle7:
 
 [@bs.set]
 external setDisplayName: (component('props), string) => unit = "displayName";
+
+[@bs.get]
+external getName: component('props) => Js.undefined(string) = "name";
+
+[@bs.get]
+external getDisplayName: component('props) => Js.undefined(string) =
+  "displayName";
```
rickyvetter, line 399: Let's use [@bs.return nullable] on these - see #526 as an example to get a better api on the Reason side. Also I think we should make the API match and just call displayName and name.

czystyl: Thanks @rickyvetter 🙌 

`[@bs.return nullable]` is nice!

PR #443, opened 2019-06-12
Update components.md

psb:
'tells the ReasonReact' -> 'tells ReasonReact'

'which takes a function and uses the return.' -> 'which takes a function and uses the return value.'


chenglou:
I see that this has already been superseded. Thanks and sorry about that!

PR #446, opened 2019-06-13
Clarifies useEffect API Documentation.

pckilgore:
I really only understood this by referencing the react-reason code itself.  I understood that `2` to mean "this is the second version of the thing we're talking about" not "call this specially named function"  (obvious in retrospect, but....).

It still took me a while to realize that I was expected to return a cleanup function from useEffect even if I didn't need to do cleanup (something I'm not used to having to do).  But I guess that begs the question how detailed ya'll want your docs to be here.

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -69,7 +69,23 @@ The component above could be called like this:
 
 The next thing you might notice looking at this example is the use of hooks (`useState`). ReasonReact binds to [all of the hooks that React provides](https://reactjs.org/docs/hooks-intro.html) with only minor API differences. Please refer to their excellent documentation for more information on how hooks work and for best practices.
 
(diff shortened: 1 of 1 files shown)


docs/components.md
```diff
@@ -69,7 +69,23 @@ The component above could be called like this:
...
+useEffect(effect, []) 
+```
+
+would translate to these two reason calls:
+
+```reason
+useEffect2(effect, (dep1, dep2))
+     /* ^^^ -- Note the number matching the depenedencies' length */
+useEffect0(effect)
+     /* ^^^ --- Compiles to javascript as `useEffect(effect, [])` */
```
rickyvetter: Thanks for this explicit callout!

rickyvetter:
Thanks for the changes. I definitely agree that this is clearer!

> It still took me a while ... return a cleanup function from useEffect

Yeah documentation here would be awesome - although you can return `None` from `useEffect` if you aren't cleaning up, so it doesn't technically need to be a function.

pckilgore:
@rickyvetter gah, just noticed a pretty horrific typo `depenedencies` => `dependencies`

pckilgore:
> although you can return `None` from `useEffect` if you aren't cleaning up, so it doesn't technically need to be a function.

Ahh, somehow missed that, thanks, much more self documenting than returning identity.

rickyvetter:
Fixed the typo. Thanks for pointing it out!

PR #448, opened 2019-06-22
ReasonReactRouter: expose function to create url type from strings

Freddy03h:
Useful for SSR.

`useUrl` take a `serverUrl` param which is an `url` type. Many SSR examples copy paste functions from ReasonReactRouter because `url` creation logics aren't exposed: [example with razzle](https://github.com/jaredpalmer/razzle/blob/eac1bd787dea392ff29a4f88dba5253e90a5f590/examples/with-reason-react/src/Router.re#L6)

### DIFF
--- a/src/ReasonReactRouter.re
+++ b/src/ReasonReactRouter.re
@@ -51,7 +51,7 @@ let safeMakeEvent = eventName =>
   };
 
 /* This is copied from array.ml. We want to cut dependencies for ReasonReact so
-    that it's friendlier to use in size-constrained codebases */
+   that it's friendlier to use in size-constrained codebases */
 let arrayToList = a => {
   let rec tolist = (i, res) =>
     if (i < 0) {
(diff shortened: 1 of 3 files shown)


src/ReasonReactRouter.rei
```diff
@@ -27,6 +27,9 @@ let unwatchUrl: watcherID => unit;
       for an example.
       */
 let dangerouslyGetInitialUrl: unit => url;
+/** expose url creation functions, usefull for SSR, create an url type from strings to give to serverUrl */
+let getUrl:
```
rickyvetter: Not sure `getUrl` makes sense when you're providing all the data. Why not `makeUrl`, `url`, or something similar?

src/ReasonReactRouter.re
```diff
@@ -133,14 +144,42 @@ type url = {
...
+    ),
+  search:
+    searchFromString(
+      switch (search) {
+      | Some(search) => search
+      | None => ""
+      },
+    ),
+  hash:
+    hashFromString(
```
rickyvetter: This is an odd choice to me - [URL](https://nodejs.org/api/url.html#url_url_hash) in JS provides a hash without the leading `#` - so using this requires some extra effort. This is especially weird for server use since we don't have this value on the server so you'd have to be constructing from some other information. Seems like bad form. But I get why this PR would have it for completeness.

src/ReasonReactRouter.re
```diff
@@ -133,14 +144,42 @@ type url = {
...
+let url = (~pathname=?, ~search=?, ~hash=?, ()) => {
+  path:
+    pathFromString(
+      switch (pathname) {
+      | Some(pathname) => pathname
+      | None => "/"
+      },
+    ),
+  search:
+    searchFromString(
```
rickyvetter: Same as below, this is weird because URL in JS provides query without the leading `?` so you have to add it yourself.

Freddy03h:
Hello, thank's for your answer!

> Folks already have access to the url type so they can construct the record by hand

That's the point, everyone had to copy-paste the functions to create the `url` type.

I'm new with Reason so it's ok if you tell me that implementation and naming had to be done an other way. I opened a PR instead of an issue first because I think code is better than a long description to start a discussion.
Maybe I came with the wrong solution, but I really want to prove the problem exist.

The situation _now_ is: on server side you receive a string url, it can be an entire url string or it can already be an object with different part (like node URL, but it can be [url-parse](https://www.npmjs.com/package/url-parse), or custom object from a framework, etc). You have to care if search/query string must contain the `?` or not (same for `#` in hash). And after you ["implement" (but in fact, copy-paste)](https://github.com/jaredpalmer/reason-react-native-web-example/blob/77f5373cee9b2044655ad138c466fc18ba2913b6/src/Routes.re#L37) the path, search/query and hash parsing in Reason to create the `url` type.

The situation I try to solve is only on the implementation part, to avoid everyone have to copy-paste function from `ReasonReactRouter`. I don't try to solve the other points.
You still have to deal on server side to create the correct `path`, `search` and `hash` strings, with the `?` and `#`. This problem already exist, and maybe we can create a larger solution but I prefer to discuss it first.

Thank you.

rickyvetter:
I guess I'm imagining some code like this (ignore the likely many errors):

```reason
module URL = {
  type t;
  [@bs.get]
  external pathname: t => string = "";
  [@bs.get]
  external search: t => string = "";
  /* realistically I don't think this even comes to the server? */
  [@bs.get]
  external hash: t => string = "";
};

/* .... */
useUrl(
  ~serverUrl={
    /* I see what you mean about logic duplication here */
    path: URL.pathname(myUrl)->Js.String.split("/", _)->Array.to_list,
    search: URL.search(myUrl),
    hash: URL.hash(myUrl),
  },
  (),
);
```

The reason I prefer this method is because if you just have the type and don't have the function, there's no assumption that any specific string encoding is correct. I (hope) that folks will take more time to make sure they are passing the correct values around.

I would much prefer this solution you've presented if there was one single known way of working with URLs in node (like `window` on the client) which we could reliably derive with only `unit` as an argument.

That said, if folks are frequently copying this code I think we should do something better for them and this approach definitely does that.

> I opened a PR instead of an issue first because I think code is better than a long description to start a discussion.

Thanks for this! I think it's much easier to discuss this topic with a concrete implementation.

@chenglou how do you feel about this?

peterpme:
Hey @Freddy03h,

In an effort to clean up this repo, I'm going to close this PR out for now. If you'd like to re-open with this updated against master, feel free.

Thanks and sorry about the delay!

PR #455, opened 2019-07-06
Fix the intro example

bitemyapp:
The examples didn't compile for me until I added this to both files.

### DIFF
--- a/docs/intro-example.md
+++ b/docs/intro-example.md
@@ -9,6 +9,8 @@ Here is a small overview of the ReasonReact API before we start. No worries if s
 ```reason
 /* file: Greeting.re */
 
+[@bs.config {jsx: 3}];
+
 [@react.component]
 let make = (~name) =>
   <button> {ReasonReact.string("Hello " ++ name ++ "!")} </button>;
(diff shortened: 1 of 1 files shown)


bitemyapp:
Okay I see what was going on here.

I had `"reason": { "react-jsx": 2 },` in my `bsconfig.json` from starting against an earlier example.

I'm assuming the docs work because they assume you'll be using a scaffold with `"reason": { "react-jsx": 3 },`.

I'm closing this as I believe the docs are working as intended but this issue is likely to tangle up other users who won't know how to solve the problem themselves.

bitemyapp:
For the sake of others who might run into this problem, the error I got was:

```
[1] >>>> Finish compiling(exit: 1)
[1] >>>> Start compiling
[1] [1/1] Building src/Index.cmj
[1] 
[1]   We've found a bug for you!
[1]   /home/callen/work/lorepub/mootapp/nu-moot/frontend/src/Index.re 3:35-42
[1]   
[1]   1 │ /* file: Index.re */
[1]   2 │ 
[1]   3 │ ReactDOMRe.renderToElementWithId(<Greeting name="John" />, "greeting");
[1]   
[1]   This function has type (~name: string) => ReasonReact.reactElement
[1]   It only accepts 1 argument; here, it's called with more.
[1]   
[1] >>>> Finish compiling(exit: 1)
```

I included the full text so that anyone googling is more likely to find this.

PR #457, opened 2019-07-17
docs: adjust some references to old record-API

lessp:
While glancing around I found some low-hanging fruit. Might be an idea to come back at another time to add some more information, but better to change them now, instead of pointing to the old API, at least!


rickyvetter:
This is awesome!  thanks.

rickyvetter:
This is awesome!  thanks.

rickyvetter:
This is awesome!  thanks.

PR #464, opened 2019-08-09
Fix cloneElement type definition

bloodyowl:


bloodyowl:
is there anything blocking this PR @chenglou @rickyvetter? right now I don't think that anybody is able to use `React.cloneElement` without `Obj.magic`.

chenglou:
Thanks for the ping. Nothing blocking, but this is a breaking change we'll have to batch with other changes.

It's breaking _unless_ you use `Obj.magic`, which we heavily discourage. The proper way is to write the specific type converter, e.g. `external littlePatch: component('props) => element = "%identity"`. This way you don't accidentally cause bigger future runtime bugs.

I know realistically most people probably just wing it with `Obj.magic`, but yeah, symbolically (and really, technically) calling this a breaking change is good.

bloodyowl:
@chenglou just realised thanks to https://github.com/reasonml/reason-react/issues/494 that the binding isn't yet released, so this can be safely merged without calling it a breaking change.

PR #469, opened 2019-08-20
Refix crossOrigin case and type

### LINKED ISSUE #468: crossorigin fix was reverted in 0.7.0
It looks like the `crossorigin -> crossOrigin` fix that was in
https://github.com/reasonml/reason-react/pull/290 (released as part of 0.6.0) was unfortunately reverted in https://github.com/reasonml/reason-react/commit/404baf2b86e2852628ccceba65287150da0189a6

~~I'll make a PR when I get a chance, if someone doesn't get to it before me.~~
Edit: made #469 to fix this

Schmavery:
Fixes #468

### DIFF
--- a/src/ReactDOMRe.re
+++ b/src/ReactDOMRe.re
@@ -275,7 +275,7 @@ type domProps = {
   [@bs.optional]
   cite: string, /* uri */
   [@bs.optional]
-  crossorigin: bool,
+  crossOrigin: string,  /* anonymous, use-credentials */
   [@bs.optional]
   cols: int,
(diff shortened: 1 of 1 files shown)


Schmavery:
Since the capitalization changes, I could also add the new named arg without removing the old one, but since we can't mark named args as deprecated (afaik) maybe it's worth waiting for the next breaking change, idk. Open to suggestion.

chenglou:
cc @rickyvetter 
thanks. I'll check whether the next version will be breaking

PR #471, opened 2019-09-02
Align forwardRef type signature with other Ref functions

sgny:
I believe the type signature for `forwardRef` should be changed. Considering the below example from [React.js documentation](https://reactjs.org/docs/forwarding-refs.html)

```
const FancyButton = React.forwardRef((props, ref) => (
  <button ref={ref} className="FancyButton">
    {props.children}
  </button>
));

// You can now get a ref directly to the DOM button:
const ref = React.createRef();
<FancyButton ref={ref}>Click me!</FancyButton>;
```
the argument that relates to `ref` in `forwardRef` should have the same type signature as the return type of `createRef`

Since `forwardRef` may also be used with `useImperativeHandle*`, this PR also updates their type signatures.

### DIFF
--- a/src/React.re
+++ b/src/React.re
@@ -61,7 +61,7 @@ module Context = {
 
 [@bs.module "react"]
 external forwardRef:
-  ([@bs.uncurry] (('props, Js.Nullable.t(Ref.t('a))) => element)) =>
+  ([@bs.uncurry] (('props, Ref.t(Js.Nullable.t('a))) => element)) =>
   component('props) =
   "forwardRef";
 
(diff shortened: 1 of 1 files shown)


rickyvetter:
I don't think this change is correct. If you were to call, for example, `<FancyButton>Click me!</FancyButton>;` (with no ref) then internally ref will be null. If we type it as `Ref.t(Js.Nullable.t('a))` then we would incorrectly allow object access on a potentially null value.

sgny:
@rickyvetter The same applies when you have `let ref = React.createRef();` but do not pass that `ref` to any component.

If you do not call `FancyButton` without a `ref`, then you have no object to access or pass to functions such as `Ref.current`, so I don't think there should be an issue.

PR #474, opened 2019-09-04
Add PupilFirst LMS to examples section

bodhish:
We use Reason-React for all of our newest front end work, and we’ve written a lot of Reason code over the course of the past year (> 30% of the total LOC now).
https://github.com/SVdotCO/pupilfirst

If you’re unfamiliar with Rails, all of the ReasonML goes in the app/javascript path.

We switched to using hooks once that was released, but we haven’t switched all of our old components to JSX3 yet - we expect that’ll happen naturally over time; the old API still works perfectly fine, and we’ve written Jsx2 compatibility modules for a few components that needed to play nicely with older ones.


chenglou:
Hey! This is great adoption; thanks. Though I'm not too sure about adding this to the examples section yet

bodhish:
@chenglou that's totally fine. :+1:  
We just want to showcase the project, this would be one of the largest, most-lived production 
@reasonml codebase.

chenglou:
Yeah sorry; I just don't want accidentally subtle promotion for the examples, disregarding how nice the product is. Plus, it's a lot to maintain if we ever release a new version, and we're already having a hard time keeping all the examples up to date.

Though on this point: would you like to write a guest blog post for ReasonReact? That's at least more professional and nobody would find it weird. No pressure!

bodhish:
That totally fine @chenglou 
We will get back to you on the guest blog part. 
Closing the issue.

PR #481, opened 2019-09-28
docs: update webpack instructions on start-page

lessp:
I've noticed there's been some confusion due to this. I hope it makes more sense to run `server` which will serve it at `localhost:xxxx`.

Hoping it's the correct file (start-page)!


chenglou:
Wow oops, thanks for catching it!

PR #487, opened 2019-10-21
docs: Update webpack script to server in installation page

g1eny0ung:
Seems that the webpack script is deprecated in development? I try reason-react today, but this makes me a little confused.

This pr is a patch of #481 .


chenglou:
Thanks!

PR #505, opened 2020-01-08
Fix method for forms in Bucklescript > 6.x

arnarthor:
We encountered a problem in bucklescript 7 where `method` is a reserved word so it would need to be entered as `method_`. This caused ReasonReact to forward it to the dom as `method_` instead of `method` despite the actual prop being named `method`

Therefor this strange fix where we alias the argument as the same thing again.


cristianoc:
https://github.com/facebook/reason/issues/2523

PR #512, opened 2020-03-03
Add bindings for useTransition hook

kiangtengl:
This PR adds bindings for the useTransition hook for React (experimental)
- useTransition
- useTransitionWithConfig // calls useTransition with the specified config

@rickyvetter  don't land until it's released in React

### DIFF
--- a/src/React.re
+++ b/src/React.re
@@ -394,6 +394,6 @@ external setDisplayName: (component('props), string) => unit = "displayName";
 [@bs.deriving abstract]
 type transitionConfig = {timeoutMs: int};
 
-[@bs.module "React"]
+[@bs.module "react"]
 external useTransition: (~config: transitionConfig=?, unit) => (callback(callback(unit, unit), unit), bool) =
   "useTransition";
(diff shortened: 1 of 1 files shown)


src/React.re
```diff
@@ -390,3 +390,10 @@ external useImperativeHandle7:
 
 [@bs.set]
 external setDisplayName: (component('props), string) => unit = "displayName";
+
+[@bs.deriving abstract]
+type transitionConfig = {timeoutMs: int};
+
+[@bs.module "React"]
```
rickyvetter: Lowercase here.

src/React.re
```diff
@@ -390,3 +390,10 @@ external useImperativeHandle7:
 
 [@bs.set]
 external setDisplayName: (component('props), string) => unit = "displayName";
+
+[@bs.deriving abstract]
```
rickyvetter: Note here for me mostly - drop this if we end up bumping BS min version to record as obj support.

peterpme:
Hey @kiangtengl 

If you don't mind, can you please update these changes and we can merge this in, thank you!

PR #519, opened 2020-04-16
Use an external for ReactDOMRE.Style.combine and remove Obj.magic

bloodyowl:


sgny:
I've been waiting for #437 to be merged for a while, as we already implemented something similar for `reason-react-native`.

That one uses the same definition for combine, but does not revise `unsafeAddProp`. There is also an `unsafeAddStyle` which allows something like
```reason
unsafeAddStyle(someStyle, {"someProp1": someValue1, "someProp2": someValue2});
```
which avoids multiple object copies when there are multiple props to add.

It also allows creating a style as below
```reason
unsafeAddStyle(make(), {"someProp1": someValue1, "someProp2": someValue2});
```
but `%identity` is always available to convert such objects for use with `combine` anyway.

rickyvetter:
Hey do you mind rebasing? This looks great to me.

PR #521, opened 2020-04-16
Simplify ref type

cknitt:
@rickyvetter as discussed on Discord:

Now that we have records as objects in BS7, let's simplify the type for React refs to:

```reason
type ref('value) = {mutable current: 'value};
```

`Ref.t`, `Ref.current` and `Ref.setCurrent` can be deprecated and removed in a later version.

### DIFF
--- a/src/React.re
+++ b/src/React.re
@@ -21,14 +21,20 @@ external createElementVariadic:
   (component('props), 'props, array(element)) => element =
   "createElement";
 
-module Ref = {
-  type t('value);
+type ref('value) = {mutable current: 'value};
 
-  [@bs.get] external current: t('value) => 'value = "current";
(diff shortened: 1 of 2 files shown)


flash-gordon:
Just two lines clarifying what the upgrade process looks like:

### Accessing current value
Before:
```reason
reactRef->React.Ref.current->Js.Nullable.toOption
```
After
```reason
reactRef.current->Js.Nullable.toOption
```

### Type aliases
Before:
```reason
type aliasedRef = React.Ref.t(Js.nullable(Dom.element));
```
After:
```reason
type aliasedRef = React.ref(Js.nullable(Dom.element));
```

### Updating ref value
Before:
```reason
React.Ref.setCurrent(elRef, None);
```

After:
```reason
elRef.current = None;
```

I also sent a doc patch.

PR #522, opened 2020-04-17
[@bs.return nullable] to simplify using Ref.t

sgny:
I meant to prepare something similar to this PR for a while now, but @cknitt's PR #521 allows avoiding another type for `ref` so this would build on that. With #521, we'd have `type t('value) = ref(Js.Nullable.t('value))` in module `Ref`.

In brief, having to deal with both `ref`s for persisted values (i.e. those created with `useRef(someValue)`) and `ref`s to components makes it painful to handle the latter. `someRef->Ref.current->Js.toOption` is hardly pleasant.

If `ref`s are defined as a record, we can repurpose existing functions to allow simpler access to component `ref`s, allowing to use `someRef->Ref.current` instead. I don't suppose setting such a `ref` to `null` is necessary, but that can be handled with a suitable function, as in `setNull`.

With this definition, below statements are both valid for getting a reference to a component:
```reason
someRef->React.Ref.current;
someRef.current->Js.toOption;
```
and likewise for setting a value 
```reason
someRef->React.Ref.setCurrent(someValue);
someRef.current = Js.Nullable.return(someValue);
```
As `useRef` returns `ref(Js.Nullable.t('value))`, this PR introduces some pain in using persisted values (needing to unwrap an `option` and wrapping a value in `Js.nullable`), but `useRef` could (and should) be defined as in #521 to return `ref('value)`.

### DIFF
--- a/src/React.re
+++ b/src/React.re
@@ -22,13 +22,14 @@ external createElementVariadic:
   "createElement";
 
 module Ref = {
-  type t('value);
+  type t('value) = {mutable current: Js.nullable('value)};
 
-  [@bs.get] external current: t('value) => 'value = "current";
+  [@bs.get] [@bs.return nullable] external current: t('value) => option('value) = "current";
   [@bs.set] external setCurrent: (t('value), 'value) => unit = "current";
+  [@bs.set] external setNull: (t('value), [@bs.as {json|null|json}] _) => unit = "current";
 };
 
-[@bs.module "react"] external createRef: unit => Ref.t(Js.nullable('a)) = "createRef";
+[@bs.module "react"] external createRef: unit => Ref.t('a) = "createRef";
 
 module Children = {
   [@bs.module "react"] [@bs.scope "Children"] [@bs.val]
--- a/src/React.re
+++ b/src/React.re
@@ -21,6 +21,8 @@ external createElementVariadic:
   (component('props), 'props, array(element)) => element =
   "createElement";
 
+type ref('value) = {mutable current: 'value};
+
 module Ref = {
   type t('value) = {mutable current: Js.nullable('value)};
 
@@ -307,7 +309,7 @@ external useCallback7:
 
 [@bs.module "react"] external useContext: Context.t('any) => 'any = "useContext";
 
-[@bs.module "react"] external useRef: 'value => Ref.t('value) = "useRef";
+[@bs.module "react"] external useRef: 'value => ref('value) = "useRef";
 
 [@bs.module "react"]
 external useImperativeHandle0:
--- a/src/ReactDOMRe.re
+++ b/src/ReactDOMRe.re
@@ -90,7 +90,7 @@ type domRef;
 
 module Ref = {
   type t = domRef;
-  type currentDomRef = React.Ref.t(Js.nullable(Dom.element));
+  type currentDomRef = React.Ref.t(Dom.element);
   type callbackDomRef = Js.nullable(Dom.element) => unit;
 
   external domRef: currentDomRef => domRef = "%identity";


src/React.re
```diff
@@ -22,13 +22,14 @@ external createElementVariadic:
   "createElement";
 
 module Ref = {
-  type t('value);
+  type t('value) = {mutable current: Js.nullable('value)};
```
cknitt, line 25: I don't think this will work. First of all it will break all existing code that does

```reason
let myRef = React.useRef(Js.Nullable.null);

<SomeComponent ref=myRef />
```

because the type of `myRef` will then be `Js.nullable(Js.nullable('value))`.

Second, refs not referring to a React component need not be nullable, e.g. when you do

```reason
let myRef = React.useRef("my initial string value");
```

So I would propose to keep the definition from my PR and add a module `ComponentRef` with a separate type plus your helper functions:

```reason
type ref('value) = {mutable current: 'value};

module Ref = {
   // as in my PR for compatibility
};

module ComponentRef = {
  type t('value) = ref(Js.nullable('value));

  // plus your helper functions
}
```

sgny: I'm aware it's a breaking change which is why I did not bring this up earlier. I already had in mind a second type like that, but I also didn't want to be stuck with `React.ComponentRef.t` (which might mean another migration when `React.Ref` disappears) or `React.Ref.componentRef`. That's why your PR simplifies things, at least, IMO. If that is not desired, I'll revise to add a second type or module. 

Once your PR is merged and `useRef` returns `ref('a)`, however, your example will not break. Otherwise, `createRef` would serve for components and `useRef` for everything else.

rickyvetter:
What if I want a ref that's not nullable?

sgny:
Then you would use `type ref` as defined in #521.

sgny:
I pushed another PR to clarify my intention, instead of explaining it. I'm obviously open to revisions, but this PR requires rather straightforward changes to existing projects

#### **case 1**: persisted value
`let someRef = useRef(someValue)`
access as `someRef.current` instead of `Ref.current(someRef)` or `someRef->Ref.current`

#### **case 2**: component ref
`let someRef = useRef(Js.Nullable.null)`
or
`let someRef = createRef()`

remove calls to `toOption`

If that is considered too much of a breaking change, we can add functions such as `currentNullable` and `setCurrentNullable` instead.

rickyvetter:
I don't think this change makes sense for a few reasons. First - it only applies to `createRef`, which is specifically intended to be used in the class api and should really be popping up frequently in new code which is using function components. Second - it creates confusion about what the actual type of Ref is with React.ref and React.Ref.t in most libraries being aliased to each other. Third - it's really simply implemented in user land with a single module and a toOption chain that can get you this kind of behavior.

If all three of these weren't true, I think this change could be worth working on. Together, I think it means that we should be focusing on the happy path of function components and useRef. That said, I appreciate the really well thought out reasoning and intentions.

rickyvetter:
Hey I want to reopen this because I had another idea that's closely related to this and wanted to get your thoughts. What if we just removed `createRef` and ask people to create the record directly?

The primary issues with `createRef` as an api is that it generates refs with `null` as default and there's no way to change this. This new "binding" would give more functionality to `createRef`. This is where things get a bit tricky - since once you can't read a ref from a dom node that you didn't pass, you never really read a dom ref created in JS (that could be null). In this way modeling `currentDomRef` as `ref(option(Dom.element))` even without the `bs.return nullable` would be safe.

First off - does this logic seem correct? Am I missing something?

If it does how do you feel about these changes?

sgny:
Basically having people use
```reason
myRef = {current: None};
```

We could also consider adding
```reason
let createRef = () => {current: None};
```
instead of a binding to React's `createRef`.

I'm quite sure you're right that this would be fine, but I've thought of two possible issues:

• when using Reason components from JS. If I use `forwardRef` in my Reason component, then the component will be passed `{current: null}` at some point. If the ref is accessed before the component is mounted, then we might have a problem.   

• when the component is unmounted, what happens to the ref? does it become `{current: null}`?

I'll try these for a bit and try to produce errors.

rickyvetter:
Yeah both of your potential issues are real. React always sets these refs as `null`. Refs passed to DOM nodes must be modeled as `ref(Js.nullable(Dom.element))`. Thanks for bringing those up.

rickyvetter:
Closing this again for the reasons above. I don't think the devx with typing is worth potential confusion of similarly named apis.

PR #523, opened 2020-04-18
DRY for domProps and props

sgny:
This is another breaking change, but no further effort would be needed to keep `props` and `domProps` in sync.

I think such a change offers benefits to users as well. Basically, they would no longer have to think if they need `props` or `domProps`, especially if they don't use `ref`.

This version preserves type names (`props` and `domProps`), but requires updating calls to constructor functions. Type annotations are needed, but `createElement`, `createElementVariadic`, `createDOMElementVariadic` already are annotated. I believe a  text search for `props` and `domProps` constructors to replace with `makeProps` is not a huge pain, but this could be revised to minimise impact on existing projects.


sgny:
This version requires type annotations (otherwise the compiler complains about weak types). For an alternative which only requires replacing calls to `domProps` with `DomProps.props` see the [`propTypes`](https://github.com/sgny/reason-react/tree/propTypes) branch in my fork.

peterpme:
Hey @sgny 

Thanks so much for this PR! In an effort to clean up & because of conflicts, I'm going to close this one out for now.

If you'd like you can re-open with resolved conflicts 😄

PR #529, opened 2020-04-21
Added uncurried hooks

johnridesabike:
Any hook which can return a function can now return an uncurried function. Uncurried versions end with a “U” (the same convention used by Belt).

```reason
let (count, setCount) = React.useStateU(() => 0);
setCount(. a => a + 1);
```

May be especially useful since the next BuckleScript will improve uncurry support.


rickyvetter:
Hey - I'm worried about the dev experience with this where all of these additional functions will show up in autocomplete and might cause additional confusion beyond the useEffectN style calls. I know this convention was taken from Belt and I don't really want to go against that, but how do you feel about `React.Uncurried.useState` etc?

johnridesabike:
That makes sense to me. I’m happy to change it 👍

johnridesabike:
They should all be namespaced under `React.Uncurried` now.

```reason
let (count, setCount) = React.Uncurried.useState(() => 0);
setCount(. a => a + 1);
```

src/React.re
```diff
@@ -394,5 +394,76 @@ external useImperativeHandle7:
...
+  [@bs.module "react"]
+  external useState:
+    ([@bs.uncurry] (unit => 'state)) =>
+    ('state, (. ('state => 'state)) => unit) =
+    "useState";
+
+  [@bs.module "react"]
+  external useReducer:
+    ([@bs.uncurry] (('state, 'action) => 'state), 'state) =>
+    ('state, (. 'action) => unit) =
```
rickyvetter, line 407: Why just uncurry the return value? The inputs also could be uncurried and this has interesting effects. See the JS output here - https://reasonml.github.io/en/try?rrjsx=true&reason=NoAQRgzgdAtg9gEwK4BsCmACARAJzQQwGMAXLAXQCg0APYtHAO3xQyQjQCU1lD6AuChgwAKUJChIGhJDhwBPMiOEByCMXx0ANBmVFiASzgMAlBgC8APh1qNaY9tXq6py4JGPb24VB17DJ8ytJfWIXNyw2Tm4kXhwsAG4KCjFoeGR0bDw9cipaeiYWSK4eegBVASFvER8PLV8SfxcrWrQHG2dA93bW6vqDIybWBhCwoQj2Ypj6BKT0Ygw1HH0GAHMAFQALNAYASQZ5swXiHE6Abzc5jAAzODg+I6XV8wfEoX1Om7hEgF9EikvFst1ltdvtSs8qottPp7stQmc3AB6ABUGAYcDoGAAtEN8GAMsQ4AsAA4oEIYYgbfQQa6SBpGDBbPCM+iYZGItyLH6zNDzT4QwbnIRFaKxYSA1abbZ7YjaLAAdzg8qwxlerAmorKkOOQKloOIpTliuVqoovwoQA

johnridesabike: Interesting. It looks like it may be related to `[@bs.uncurry]`?

I don’t think we need to uncurry the whole `useReducer` hook to eliminate the runtime uncurrying, just the reducer argument:

```
[@bs.module "react"]
external useReducer:
  ((. 'state, 'action) => 'state, 'state) => ('state, (. 'action) => unit) =
  "useReducer";
```

https://reasonml.github.io/en/try?rrjsx=true&reason=NoAQRgzgdAtg9gEwK4BsCmACARAJzQQwGMAXLAXQCg0APYtHAO3xQyQjQCU1lD6AuChgwAKUJChIGhJDhwBPMiOEByCMXx0ANBmVFiASzgMAlBgC8APh1qNaY9tXq6py4JGPb24VB17DJ8ytJfWIXNyw2Tm4kXhwsAG4KCjFoeGR0bDw9cipaeiYWSK4eegBVASFvER8PLV8SfxcrWrQHG2dA93bW6vqDIybWBhCwoQj2Ypj6BKSU2ERUTFwCEhyaOkZmVgno2NKAJgqlGu6HPwHOlranO06VU97dBovLIZHzcKLd6cSKdGIMGocPoGABzAAqAAs0AwAJIMAFmQHEHCdADebn+GAAZnA4HxkcCweZCYkhPpOri4IkAL6-LFAkEQ6FwhGlD6VHxA7T6Akg0KdAD0ACoMAw4HQMABaIb4MAZYhwQEABxQIQwxEh+ggOMkzwYGGheEN9EwwsFbiEVutNttQPpaABVJJwkGGKEXxKOGEjLBUJh8OI2iwAHc4CGsMYydsol7SlVfcyA2zg2GI1G3J6pjgDj6UUz-aziKVU+HI7TEkA

yawaramin: Hmm I would actually recommend to keep `[@bs.uncurry]` there. In Ricky's example, BuckleScript is unable to uncurry `stringThenInt` fully because it contains a closure. But if you turned it into `let stringThenInt = (str, _i) => str;`, which doesn't create a new closure, then BuckleScript can uncurry it just fine. And in fact this form is what people will normally use.

johnridesabike: I think @yawaramin makes a good point, but I’m willing to go with whatever the reason-react maintainers think is the best choice.

sgny:
@johnridesabike

I haven't managed to figure out how to use the function returned by `useState` as in your example above.

```reason
let (count, setCount) = React.Uncurried.useState(() => 0);
setCount(. a => a + 1);
```
Compiler returns:
```
This is an uncurried BuckleScript function. It must be applied with a dot.

Like this: foo(. a, b)
Not like this: foo(a, b)
```
I've been using version 7.2.2 of the compiler, but I've tried with 5.0.3 and 7.3.1 as well. What am I missing?

rickyvetter:
@sgny I've noticed the same thing. It appears to be related to the location of the calling function. Specifically it errors for me with your message when used at the toplevel.

Error:
```reason
[@bs.module "react"]
external useState:
  ([@bs.uncurry] (unit => 'state)) => ('state, (. ('state => 'state)) => unit) =
  "useState";

let (count, setCount) = useState(() => 0);
setCount(. a => a + 1);
```

No error:
```reason
let bar = () => {
  let (count, setCount) = useState(() => 0);
  setCount(. a => a + 1);
};
```
```reason
let (count, setCount) = useState(() => 0);
let _ = setCount(. a => a + 1);
```

I suspect it's because the . annotation does not get applied correctly in Pstr_eval (toplevel execution). I'll open an issue in BuckleScript.

sgny:
@rickyvetter 
Thanks for the clarification. I suppose I wouldn't have run into it if I were using an actual component and using the hook inside that.

src/React.re
rickyvetter: Switching to dot notation from `bs.uncurry` essentially makes my weird version of `stringThenInt` impossible to write. The idea being that (at least in JS in the context of a function we know will get called with two arguments) this split version of `stringThenInt` is always less performant and shouldn't be encouraged. If we are interested in adopting the very strict dot notation return values for performance then I think we should be interested in enforcing inputs as well. That said - I'm not confident that performance is the reason folks are interested in changing the return value. @johnridesabike can you explain a bit why you're interested in this PR?

johnridesabike: Without an uncurried return value, there’s no way to avoid a `Curry._*` function at runtime when using a hook’s dispatch. Even though uncurried functions haven’t been heavily used in the past, BuckleScript 7.3 improves uncurry support so people can use dot-notation functions in almost all cases when they could also use regular functions.

Using guaranteed-uncurried functions increases performance, reduces bundle size, and makes JavaScript more readable (all by avoiding the `Curry` module). But I think the important part of this PR is offering users a choice. Even though many users may not need to care about runtime uncurrying, those that do care (for any of the aforementioned benefits) will appreciate the option to avoid it. 

The choice part is why I feel skeptical that dot-notation should be required for the input. If a user is interested in improving performance and JS readability, they are probably able to write a reducer function that won’t cause the `[@bs.uncurry]` overhead. I don’t think we need to force maximum performance, just allow the option for those who want it.

That said, I don’t feel strongly *against* using a dot-notation reducer as input either. It just doesn’t seem necessary.

yawaramin: I like the ergonomics of `[@bs.uncurry]` but it is reasonable to expect people to use uncurried syntax if they're specifically using the `React.Uncurried` module. So @rickyvetter it's your call 🙂 If we do go with dot-syntax though I would recommend using it consistently in all the uncurried bindings. Another thing to think about is if we want to take advantage of BS 7.3's improved support e.g. by using labelled parameters. Maybe not right now though.

rickyvetter:
This isn't a breaking change and I want to make sure we get this api right. Not going to push for 0.8.0 release. Still want to try some things to check differences with full dot notation.

jfrolich:
The `dispatch` function identity is guaranteed to be stable in React. So the curried version actually changes the behavior of this hook in unexpected ways. So I wonder if we should make the uncurried the default at some point? (at least this should be clearly documented IMO).

rickyvetter:
@jfrolich the identity of the dispatch function is stable in both the curried and uncurried version. I don't think this comes into play.

jfrolich:
> @jfrolich the identity of the dispatch function is stable in both the curried and uncurried version. I don't think this comes into play.

🙈Of course, I overlooked something.

PR #536, opened 2020-04-27
memoize callbackDomRef

sgny:
Callback refs may currently be created with `callbackDomRef` and memoized with `useCallback` as `r->callbackDomRef->useCallback`. This PR, calls `useCallback` instead of using `%identity` for typecasting and achieves the same result with `r->callbackDomRef`, without breaking the API. As callback refs are supposed to be called only when mounting and unmounting, `[]` is the right dependency.

If there are any concerns of unnecessary overheads due to `useCallback`, An alternative may be to add something like `callbackDomRefMemoized`, but perhaps it would be better to add documentation on ref best practices instead.

### DIFF
--- a/src/ReactDOMRe.re
+++ b/src/ReactDOMRe.re
@@ -94,7 +94,15 @@ module Ref = {
   type callbackDomRef = Js.nullable(Dom.element) => unit;
 
   external domRef: currentDomRef => domRef = "%identity";
-  external callbackDomRef: callbackDomRef => domRef = "%identity";
+
+  [@bs.module "react"]
+  external callbackDomRef:
+    (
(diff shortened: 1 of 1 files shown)


rickyvetter:
I feel like there is too much magic happening here. It appears to be a non-hook casting operation and actually comes with a heavy runtime choice. The perf is one aspect - but what if someone ends up calling this cast outside of a component render cycle and hits runtime errors? When they read source things will look very strange.

Much more into `callbackDomRefMemoized`, but personally I'd like to get rid of `domRef` and `callbackDomRef` entirely with some additional work on the domProps type. I think this is the cleanest and most straightforward way to model this. IMO the end-state api looks something like:

```reason
<div ref={`callback(r->useCallback)} />
```

PR #537, opened 2020-04-27
Add GraphQL & Apollo example

dylanirlbeck:
It was my opinion that you can't do React, and in particular ReasonReact, without talking about GraphQL. Here's an example using `graphql-ppx` and Apollo Client together!

I'm definitely open to feedback, and thanks in advance!

cc @peterpme


website/sidebars.json
```diff
@@ -1,10 +1,6 @@
 {
   "docs": {
-    "Getting Started": [
-      "what-and-why",
-      "installation",
-      "intro-example"
-    ],
+    "Getting Started": ["what-and-why", "installation", "intro-example"],
```
peterpme, line 3: can you undo the prettier here?

docs/graphql.md
```diff
@@ -0,0 +1,87 @@
+---
+title: GraphQL
```
peterpme, line 2:
```suggestion
title: Apollo & GraphQL
```

PR #539, opened 2020-04-28
Remove compilation warnings

bloodyowl:


bloodyowl:
@peterpme might sneak this in before 0.8.0 so that people running bsb don't have a compilation warning

peterpme:
I'll see what I can do :smile:

PR #541, opened 2020-04-28
Add back old bs.obj syntax

### LINKED ISSUE #540: bsrefmt version removes bs.obj externals
Was testing with @peterpme and at least some projects won't be able to build if we use bsrefmt that we are pinned to now. https://github.com/reasonml/reason-react/commit/abf61ee706c8b4b51172598c8a6c409716100903 for a manual fix. cc @bloodyowl as it's related to https://github.com/reasonml/reason-react/pull/539

bloodyowl:
Fixes #540

### DIFF
--- a/src/React.re
+++ b/src/React.re
@@ -96,7 +96,7 @@ external memoCustomCompareProps:
 module Fragment = {
   [@bs.obj]
   external makeProps:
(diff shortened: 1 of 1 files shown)


rickyvetter:
Ahh yeah. good catch.

PR #542, opened 2020-04-28
Add additional explanations to first example.

yunti:
As discussed on discord, some small brief changes for clarity.

### DIFF
--- a/docs/simple.md
+++ b/docs/simple.md
@@ -4,7 +4,9 @@ title: A List of Simple Examples
 
 ### A Basic Greeting Component
 
-Reason's returns are implicit so you don't need to write `return`, it'll be the last item in the block:
+Reason's returns are implicit so you don't need to write `return`, it'll be the last item in the block.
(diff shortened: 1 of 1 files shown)


docs/simple.md
```diff
@@ -4,7 +4,9 @@ title: A List of Simple Examples
 
 ### A Basic Greeting Component
 
-Reason's returns are implicit so you don't need to write `return`, it'll be the last item in the block:
+Reason's returns are implicit so you don't need to write `return`, it'll be the last item in the block.
```
peterpme, line 7: Some things to note:
- Reason's `return` statements are implicit so you don't need to write `return`. It'll always be the last item in the block
- Reason has labelled parameters (arguments) that are prefixed by a tilde, eg: `~message`
- Since everything in ReasonReact is typed, we need to wrap our message in `React.string(message)`

peterpme:
@yunti thank you! lmk what you think of my comment :smile:

yunti:
@peterpme. Perfect! It’s hard keeping it brief but still covering the essentials.  Your revision is spot on.

PR #544, opened 2020-04-28
Docs: add information about using useState with form events

### LINKED ISSUE #476: Documentation: add info about using `useState` with `ReactEvent` target values
This seems to be a common pitfall among developers converting from ReactJS. In JavaScript, you can do the following:
```js
const [count, setCount] = useState(0);
// later in the component:
setCount(event.target.value);
```

When writing the same component in Reason, developers tend to incorrectly write it like this:
```re
let (count, setCount) = React.useState(_ => 0);
// later in the component:
setCount(_ => ReactEvent.Form.target(event)##value);
```

This raises an error from React, and the state will fail to update correctly. The issue is that `setState` dispatches are executed after React has cleaned up the synthetic event. Because the event target's value is accessed within a function, the event doesn't exist anymore by the time the function executes.

Correct code:
```re
let (count, setCount) = React.useState(_ => 0);
// later in the component:
let newCount = ReactEvent.Form.target(event)##value;
setCount(_ => newCount);
```

I'm happy to submit a PR with clarifying information. Is there a preference on which part of the docs it should go in? E.g.: in the Components page, on a page under ReactJS Idioms Equivalents, or somewhere else?

johnridesabike:
Explains how to correctly use `useState` dispatches with form event values. Fixes #476


docs/usestate-event-value.md
```diff
@@ -0,0 +1,75 @@
+---
+title: Using event values with useState
```
peterpme, line 2:
```suggestion
title: Using Event Values with useState
```

peterpme:
Thank you @johnridesabike :smile:

PR #550, opened 2020-04-29
readme updates

peterpme:
The goal of this README is for a newcomber to feel confident with using Reason and knowing there is a community behind it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# [ReasonReact](https://reasonml.github.io/reason-react/) - ReasonML bindings for React.js
+# [ReasonReact](https://reasonml.github.io/reason-react/) - ReasonML / BuckleScript bindings for React.js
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,41 +1,82 @@
...
+  "re:build": "bsb -make-world -clean-world",
+  "re:watch": "bsb -make-world -clean-world -w"
+}
+```
+
+Copy the `bsconfig.json` file from our docs located [here](https://reasonml.github.io/reason-react/docs/en/installation#adding-reason-to-an-existing-reactjs-project-create-react-app-nextjs-etc)
+
+## Using Your Favorite Javascript Libraries
+
+The same way that Tyepscript has `type annotations`, we have `bindings`. Bindings are libraries that allow you to import a popular project (like lodash) or to import your own local file. ReasonReact is in fact an example of a binding!
```
flash-gordon, line 70: Typo in TypeScript :)

Issue #530, opened 2020-04-21
Avoid curry for dispatch returned by useReducer


Austaras:
I have following code 
```reason
[@react.component]
let make = () => {
  let (_, dispatch) = React.useReducer((state, _) => state + 1, 0);
  <button onClick={_e => dispatch(1)}> {ReasonReact.string("+")} </button>;
};
```
which compiles to
```js
function App(Props) {
  var match = React.useReducer((function (state, param) {
          return state + 1 | 0;
        }), 0);
  var dispatch = match[1];
  return React.createElement("button", {
              onClick: (function (_e) {
                  return Curry._1(dispatch, 1);
                })
            }, "+");
}
```
Curry for dispatch is not needed, can this be fixed or it's limited by something else?

Austaras:
As my limited knowledge for bucklescript, maybe dispatch should be typed as `([@bs] 'action => unit)`

johnridesabike:
This PR would address this issue #529

Austaras:
> This PR would address this issue #529

cool, hope your pr will be merged soon

johnridesabike:
#529 got messed up. It was merged by accident, but none of its code was actually added to the master branch. #551 re-opened it.

PR #551, opened 2020-04-29
Add uncurried hooks

johnridesabike:
```reason
let (count, setCount) = React.Uncurried.useState(() => 0);
setCount(. a => a + 1);
```

The original pull request for this got messed up, so this is re-opening it.

Original PR: #529

Fixes: #530


rickyvetter:
I think I was getting a little aggressive here. Since they are externals bs.uncurry should be enough. Please swap reducer back to [@bs.uncurry] and then rebuild to resolve conflicts and we can get this in!

src/React.re
```diff
@@ -419,6 +419,76 @@ external useImperativeHandle7:
...
+module Uncurried = {
+  [@bs.module "react"]
+  external useState:
+    ([@bs.uncurry] (unit => 'state)) =>
+    ('state, (. ('state => 'state)) => unit) =
+    "useState";
+
+  [@bs.module "react"]
+  external useReducer:
+    ((. 'state, 'action) => 'state, 'state) => ('state, (. 'action) => unit) =
```
rickyvetter, line 431: Swap this input back.

src/React.re
```diff
@@ -419,6 +419,76 @@ external useImperativeHandle7:
...
+    "useReducer";
+
+  [@bs.module "react"]
+  external useReducerWithMapState:
+    (
+      [@bs.uncurry] (('state, 'action) => 'state),
+      'initialState,
+      'initialState => 'state
```
rickyvetter, line 439: Unrelated to your PR but I think this is unsafe. Should probably add bs.uncurry here as well.

src/React.re
```diff
@@ -419,6 +419,76 @@ external useImperativeHandle7:
...
+    ) =>
+    ('state, (. 'action) => unit) =
+    "useReducer";
+
+  type callback('input, 'output) = (. 'input) => 'output;
+
+  [@bs.module "react"]
+  external useCallback:
+    ([@bs.uncurry] ('input => 'output)) => callback('input, 'output) =
```
rickyvetter, line 448: Pretty sure you want to do all of the `callback` type returns as well = https://reasonml.github.io/try?rrjsx=true&reason=C4TwDgpgBAxghgGwQIzjA1gCgOQEsA0U2A9gJRQC8RulAfEcQNwBQzA2gALIDOAdALbEAJgFcE0AEQAnCGmASAusyhQIAD2AQpAO0RQR3CAGFEKNOgBcylVEycevEdpgipUkAtt5tYEcDoMfr7ApOQU9PBIqBg4uD5+hCRBfmHWKhIGxqbR6BIszOL+AGbExJS2YfQA3gUQ-obARsRO-lSZJlHmmJiVUAAMpCwNTS09LAC+jEA

src/React.re
johnridesabike: Do you think we should go ahead and add `[@bs.uncurry]` to that argument on the regular `useReducerWithMapState` too?

src/React.re
johnridesabike: All of the `Uncurried.useCallback*` functions should return type `Uncurried.callback`, defined as `type callback('input, 'output) = (. 'input) => 'output;`

https://reasonml.github.io/try?rrjsx=true&reason=C4TwDgpgBAxghgGwQIzjA1gCgOQEsB2YArsADRTYD2JxwAlFALxSYB0FBtDjAfBdcFoBuAFAiA2gAFkAZ1YBbSgBMiCaACIAThDTB1AXREQAHsAib8iKERkQAwohRp0ALhFQWU2ayL4YRTU0QfRY8QhImPioaEjpuPngkVAwcThJyaMFYpncodRt7R2T0dVERNWAoADNKSiYWeKgAb1yKqFtgO2p8SuYChyTnTDhIqBGAaigARjpRDw6u32A2KAAGWZEAX1EgA

However I did just discover another bug. It doesn’t look like Reason is good at handling arity-zero uncurried functions. I’m not sure what the best way around that is, unless I’m missing something.

https://reasonml.github.io/try?rrjsx=true&reason=C4TwDgpgBAxghgGwQIzjA1gCgOQEsB2YArsADRTYD2JxwAlFALxSYB0FBtDjAfBdcFoBuAFAiA2gAFkAZ1YBbSgBMiCaACIAThDTB1AXREQAHsAib8iKERkQAwohRp0ALhFQWU2ayL4YRTU0QfRY8QhImPioaEjpuPngkVAwcThJyaMFYpncodRt7R2T0dVERNWAoADNKSiYWeKgAb1yKqFtgO2p8SuYChyTnTExGgAY6UQ8Ort9gNgmRAF9RIA

Seems to be related to https://github.com/facebook/reason/issues/2357

src/React.re
rickyvetter: Yes. As a separate PR that would be great.

src/React.re
rickyvetter: Oh I was oblivious. Good call for keeping continuity with the other set.

Yah that issue provides a workaround but it is kind of annoying that this error exists. I don't think it should block this PR though - if the feature is never used then a better experience won't be prioritized.

yawaramin:
@johnridesabike if #574 gets merged, your diff will shrink to only two bindings per hook :-)

peterpme:
@johnridesabike @rickyvetter all good here? Would love to get this in a state of mergeability :smile:

johnridesabike:
@peterpme I don't have anything else I'm waiting for. Merging is fine with me as long as no one else has any issues.

PR #555, opened 2020-05-01
Fix Readme Bug and add youtube Videos

peterpme:
- fix the getting started
- add links to youtube videos
- link directly to docs


peterpme:
since this is moving stuff around im gonna merge it in :smile:

PR #566, opened 2020-05-05
React.jsx related deprecations

rickyvetter:
Needs work on the ppx side as well to support fully.

The goal here is to:

1. move to `React.jsx` as the default compile target for Reason JSX. This is in line with babel transforms and the API is much simpler to bind to.
2. Deprecate as many `createElement` bindings as possible to consolidate on JSX syntax+exactly one createElement binding at `React.createElement`. To do this correctly I will also remove the `jsx` bindings in `ReactDOMRe` and fiddle with the type of DOM elements to make that work.


src/React.re
```diff
@@ -24,6 +24,18 @@ external createElementVariadic:
...
+
+[@bs.module "react"] [@ocaml.deprecated "Please use JSX syntax directly."]
+external jsx: (component('props), 'props) => element = "jsx";
+
+[@bs.module "react"] [@ocaml.deprecated "Please use JSX syntax directly."]
+external jsxs: (component('props), 'props) => element = "jsxs";
+
+[@bs.module "react"] [@ocaml.deprecated "Please use JSX syntax directly."]
+external jsxsKeyed: (component('props), 'props, string) => element = "jsxs";
+
```
rickyvetter, line 38: The idea here is to have the ppx disable these warnings. Wish we were working with OCaml 4.08 - https://github.com/ocaml/ocaml/pull/1804

rickyvetter:
I have merged jsx but not the deprecations. Need support on BuckleScript/ppx side, which needs support on a stable branch of React before we can do those.

davesnx:
Done in https://github.com/reasonml/reason-react/pull/714 <3

PR #571, opened 2020-05-06
Adds React.component as a new binding for making components

rickyvetter:


src/React.re
```diff
@@ -12,6 +12,9 @@ type componentLike('props, 'return) = 'props => 'return;
 
 type component('props) = componentLike('props, element);
 
+/* this function exists to prepare for making `component` abstract */
+external component: componentLike('props, element) => component('props) = "%identity";
```
yawaramin, line 16: Seems like this identity isn't really needed because type `component('props)` is already aliased to `componentLike('props, element)` ... so the compiler should automatically equate them.

rickyvetter: Yes not needed now. It needs to be added though so that the ppx can call it so that we can make component abstract. If we do it together then we need to manage a breaking RR and BS release simultaneously. If we do it this way we can do breaking BS then breaking RR.

PR #573, opened 2020-05-06
Docs: add explanation for how to import components with default props.

johnridesabike:
This question comes up regularly on Discord.


johnridesabike:
Shadowing `makeProps` is an interesting technique too. The compiled JS does look like it may be more lightweight too since it only adds a new function to the props, and doesn’t call `React.createElement` an extra time.

It feels a bit hacky to me though, so I was hesitant to put it in the PR. But I don’t mind adding it if we all think it’s worth mentioning.

```reason
module Greeting = {
   [@bs.module "./MyJavascriptFile.js"] [@react.component]
   external make: (~name: string) => React.element = "default";
   let makeProps = (~name="Peter", unit) => makeProps(~name, unit);
 };

let f = () => <Greeting  />
```
https://nit.sketch.sh/?code=LYewJgrgNgpgBAcQE4xgFwJYDsDmcC8cA3gFBzkDaAAgEYDOAdKJLHAEQMD0AsgJ4BSAQwBugugGMkGAA5oAYhlgMAVnTYBdONRSDxaBuJDBpILDCxp1ZcjAAeaGEiyCocYIIDWMAFxwAFAB+zsA+cHRoUrgAlAQAfHAASjC6+jCwIRYE7GAwAGaC0GhsANzWcLBobp4wAApIINJ0WYHBMPhsNeiObAA0cBBYGGgx+PHuXnUNdC2CIX0DQ1GlcAC+pSQVcLnNI-EAPMiomLjknLFAA

docs/importing-js-into-reason.md
```diff
@@ -22,3 +22,42 @@ module Greeting = {
 
 let make = () => <div> <Greeting name="Peter" /> </div>;
 ```
+
+### Default props and JavaScript components
+
+Sometimes you need your JavaScript component to have a default prop. How you
+handle this will depend on the details of the situation.
+
+#### Scenario 1: The JavaScript component prop is already coded to have a default
```
peterpme, line 31: do you mind adding a javascript example of what you mean here, ie:

```
function Greeting({ name = "Peter" }) {}
```

vs.

```
function Greeting({ name }) {}

Greeting.defaultProps = { name: "Peter" }
```

I know you probably mean the 2nd one but I think its worth being explicit, perhaps worth mentioning even version #1 (and what happens there)

docs/importing-js-into-reason.md
```diff
@@ -22,3 +22,42 @@ module Greeting = {
...
+
+Props (or any function argument) with a default is just an optional prop as far
+as its signature is concerned. To bind to it, we just make the prop optional.
+
+```reason
+module Greeting = {
+  [@bs.module "./MyJavascriptFile.js"] [@react.component]
```
peterpme, line 38:
```suggestion
  [@bs.module "./Greeting.js"] [@react.component]
```

docs/importing-js-into-reason.md
```diff
@@ -22,3 +22,42 @@ module Greeting = {
...
+
+#### Scenario 2: The JavaScript component prop doesn't have a default, but we want to add one
+
+In JavaScript, the logic for default arguments is handled at runtime inside the
+function body. If our external does not already have a default, we'll need to
+wrap it with our own component to add the logic.
+
+```reason
+module GreetingJs = {
+  [@bs.module "./MyJavascriptFile.js"] [@react.component]
```
peterpme, line 55:
```suggestion
  [@bs.module "./Greeting.js"] [@react.component]
```

docs/importing-js-into-reason.md
```diff
@@ -22,3 +22,42 @@ module Greeting = {
...
+  external make: (~name: string) => React.element = "default";
+};
+
+module Greeting = { 
+  [@react.component]
+  let make = (~name="Peter") => <GreetingJs name />;
```
peterpme, line 61: I kind of like GreetingJs as a pattern, interesting!

peterpme:
Hey thanks so much, just a few nits :smile:

docs/importing-js-into-reason.md
johnridesabike: It's been a while since I've used ReactJS, but wouldn't either JS code result in the same Reason binding?  I'm fine adding the JS example in either style (or both), but let me know if I'm missing something.

peterpme:
Thank you!

PR #577, opened 2020-05-17
RFC: Adding a ReactErrorBoundary component to deal with errors

bloodyowl:
Given there's no API to bind to classes, I think it could be nice to
provide a lightweight ErrorBoundary component within the bindings.

### DIFF
--- a/src/ReactErrorBoundary.re
+++ b/src/ReactErrorBoundary.re
@@ -1,0 +1,45 @@
+type reactComponentClass;
+
+[@bs.module "react"] external component: reactComponentClass = "Component";
+
+type info = {componentStack: string};
+
+type params('error) = {
+  error: 'error,
+  info,
+};
+
+let getErrorBoundary = [%bs.raw
+  {|
+  function (Component) {
+    function ErrorBoundary(props) {
+      Component.call(this);
+      this.state = {error: undefined};
+    };
+    ErrorBoundary.prototype = Object.create(Component.prototype);
+    ErrorBoundary.prototype.componentDidCatch = function(error, info) {
+      this.setState({error: {error: error, info: info}})
+    };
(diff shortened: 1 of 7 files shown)


src/ReactErrorBoundary.re
```diff
@@ -0,0 +1,45 @@
+type reactComponentClass;
+
+[@bs.module "react"] external component: reactComponentClass = "Component";
```
bloodyowl, line 3: This is to ensure the `React` import matches the `bsconfig` spec for the project.

peterpme:
This is cool! Any reason why we can't use the native ErrorBoundary that ships with React?

Also, in an effort to get better at documentation, mind putting together a simple example under "ReactJS Idioms"?

bloodyowl:
@peterpme AFAIK React doesn't export an `ErrorBoundary` component, it just gives the ability to use the `componentDidCatch` lifecycle method. Did I miss something?

Added some documentation there: https://github.com/reasonml/reason-react/pull/577/commits/9bf599ff05a1766636c02bc3c35928ea0486faa7

rickyvetter:
Can you rename as `ReasonReactErrorBoundary`? Going forward anything prefixed with `React` is reserved for 0-runtime. I also want to signal that this will be deprecated as soon as function component support for catch is provided. So if we can signal that in docs/source that would be great. Idea makes a lot of sense though for sure.

bloodyowl:
@rickyvetter Done. Do you happen to know anything about an alternative functional API in React for error boundaries? I can't seem to find anything on the issue tracker.

rickyvetter:
Thanks - this looks good! Will merge with a bunch of other 0.9 stuff soon. I don't know of anything aside from proposals in the RFC repo and I didn't see any activity from the core team there.

Issue #579, opened 2020-05-18
Better ReasonReactRouter server support


rickyvetter:
url, path, hash, and search should all take an optional raw string parameter for server usage. To be able to construct a URL for useUrl using the exact same logic that the client path will.

davesnx:
Fixed in https://github.com/reasonml/reason-react/pull/580

PR #580, opened 2020-05-18
Allow ReasonReactRouter.url creation from string

EnriqueVidal:
- [x] Add basic parser for path
- [x] Add basic parser for hash
- [x] Add basic parser for search
- [x] Add `fromServer` function that takes a string and return `url`

Rel: #579


src/ReasonReactRouter.re
```diff
@@ -67,50 +67,68 @@ let arrayToList = a => {
...
-  switch ([%external window]) {
-  | None => ""
-  | Some((window: Dom.window)) =>
-    switch (window |> location |> search) {
-    | ""
-    | "?" => ""
-    | raw =>
-      /* remove the preceeding ?, which every search seems to have. */
-      raw |> Js.String.sliceToEnd(~from=1)
+let hash = (~serverUrlString=None, ()) =>
```
rickyvetter, line 109: Hash isn't sent to the server so no point in adding the overhead for support here.

src/ReasonReactRouter.re
```diff
@@ -67,50 +67,68 @@ let arrayToList = a => {
...
-  | Some((window: Dom.window)) =>
-    switch (window |> location |> hash) {
-    | ""
-    | "#" => ""
-    | raw =>
-      /* remove the preceeding #, which every hash seems to have.
-         Why is this even included in location.hash?? */
-      raw |> Js.String.sliceToEnd(~from=1)
+
+let path = (~serverUrlString=None, ()) =>
```
rickyvetter, line 92: This pattern is a bit odd. Just go for ~serverUrlString=?.

src/ReasonReactRouter.re
```diff
@@ -145,13 +163,16 @@ let urlNotEqual = (a, b) => {
         listNotEqual(aRest, bRest);
       }
     };
-  };
   a.hash !== b.hash || a.search !== b.search || listNotEqual(a.path, b.path);
 };
 type watcherID = unit => unit;
-let url = () => {path: path(), hash: hash(), search: search()};
+let url = (~serverUrlString=?, ()) => {
+  path: path(~serverUrlString, ()),
```
rickyvetter, line 170: ~serverUrlString?

src/ReasonReactRouter.re
```diff
@@ -166,7 +187,7 @@ let unwatchUrl = watcherID =>
   | Some((window: Dom.window)) =>
     removeEventListener(window, "popstate", watcherID)
   };
-
+let fromServer = serverUrlString => url(~serverUrlString, ());
```
rickyvetter, line 190: Don't expose this.

src/ReasonReactRouter.re
```diff
@@ -145,13 +163,16 @@ let urlNotEqual = (a, b) => {
...
 type watcherID = unit => unit;
-let url = () => {path: path(), hash: hash(), search: search()};
+let url = (~serverUrlString=?, ()) => {
+  path: path(~serverUrlString, ()),
+  hash: hash(~serverUrlString, ()),
+  search: search(~serverUrlString, ()),
+};
 /* alias exposed publicly */
-let dangerouslyGetInitialUrl = url;
+let dangerouslyGetInitialUrl = () => url();
```
rickyvetter, line 175: Pass `serverUrlString` here.

src/ReasonReactRouter.re
```diff
@@ -67,50 +67,68 @@ let arrayToList = a => {
...
-      /* remove the trailing /, which some pathnames might have. Ugh */
-      let raw =
-        switch (Js.String.get(raw, Js.String.length(raw) - 1)) {
-        | "/" => Js.String.slice(~from=0, ~to_=-1, raw)
-        | _ => raw
-        };
-      raw |> Js.String.split("/") |> arrayToList;
-    }
+let hsMatcher = [%re "/(#[^?&]+)?([?&].*)?$/"];
+let pathMatcher = [%re "/(#|\\?).+$/"];
```
rickyvetter, line 71: Can we just split on `?` and send these separate ways? Did you write these regex from scratch or get them from somewhere?

src/ReasonReactRouter.re
EnriqueVidal: Oh gotcha, less code to maintain :)

src/ReasonReactRouter.re
EnriqueVidal: Oh sure, I used that pattern to not have to check inside of `url()` but yeah I switch it to `?`

src/ReasonReactRouter.re
EnriqueVidal: will do.

src/ReasonReactRouter.re
EnriqueVidal: ahhh, okay I'll use this instead of `fromServer` also, will update tests to call this instead.

src/ReasonReactRouter.re
EnriqueVidal: I wrote them from scratch, we can split on ?since we're not dealing with hash anymore; My idea was that since I can trust a string I wanted to account for the possibility of things like hash and search being mixed in, I will ammed your comments and we can take another look.

peterpme:
Hey @EnriqueVidal thank you so much for this! After you get through Ricky's review, do you mind adding some docs around this as well? Thanks!

EnriqueVidal:
@peterpme absolutely :) be warned, english is definitely not my native language so there may be a bunch of typos, will try my best though.

peterpme:
No worries, I'll help :smile: Thanks!

src/ReasonReactRouter.re
```diff
@@ -145,11 +160,18 @@ let urlNotEqual = (a, b) => {
         listNotEqual(aRest, bRest);
       }
     };
-  };
   a.hash !== b.hash || a.search !== b.search || listNotEqual(a.path, b.path);
 };
 type watcherID = unit => unit;
-let url = () => {path: path(), hash: hash(), search: search()};
+let url = (~serverUrlString=?, ()) =>
+  switch (serverUrlString) {
```
EnriqueVidal, line 167: @rickyvetter I thinkered with this, but I can't find a way to do this without cheking `~serverUrlString` like this, perhaps there's some reason featured I am not aware of, I know about `Belt`, but since I didn't see it used anywhere I thought this was a design choice, so for now I've added a switch to check what `~serverUrlString`  is before returning a new `url`, please let me know if there's any way to improve this and I'll get back to working on it.

EnriqueVidal:
@peterpme I see there are two markdowns inside of docs for router, which is the one that needs to be updated?

EnriqueVidal:
@rickyvetter Sorry to mention you directly, am I missing anything else from the previous comments?

rickyvetter:
Hey sorry for the delay here. Changes look great. Thank you!

PR #582, opened 2020-05-19
fix syntax for desugared component demo

fakenickels:
If people would try run the snippet they would get a syntax error because variables can't be capitalized


anmonteiro:
I think this is on purpose. The snippet isn't meant to be run, but instead give an example of what the PPX generates. 

I think this was discussed in another issue / PR but I couldn't find it.

rickyvetter:
Yeah this is intentional. `props` is not accessible if you use the ppx and it's done by uppercasing the variable in the ast. There are actually a number of additional transforms we do to the component that are excluded in these docs because they might be a little overwhelming. Makes me wonder if they are worth addressing or not.

Not planning to change it. Open to including a note that this is an intentional error though.

PR #590, opened 2020-05-27
[draft] 0.10.0 deprecations

bloodyowl:
### Remove `ReactEventRe`

Every bit of the module has been deprecated for a while now, I think we can safely remove it.

### Deprecate old API and compat layer

They won't be there in 1.0.0.

### Deprecate old APIs in ReactDOMRe

- `findDOMNode`: going to be removed, as there's no need with React function API
- `domElementToObj`: not needed anymore I guess? Only recommended usage was in the deprecated event API
- `renderToElementWithClassName`, `renderToElementWithId`, `createRootWithClassName`, `createRootWithId`, `hydrateToElementWithClassName`, `hydrateToElementWithId`: removed in favour of `React.querySelector(x)` with a `switch` or an `Option.map`

### Deprecate old namespace

Those should be pretty easy to search & replace:

- `ReasonReact.string` -> `React.string`
- `ReasonReact.array` -> `React.array`
- `ReasonReact.null` -> `React.null`
- `ReasonReact.refToJsObj` -> No need?


bloodyowl:
@rickyvetter Maybe while `React.jsx` isn't released we can update the ppx:

- `ReactDOMRe.createElement(x, ...)` -> `React.createElement(ReactDOM.stringToComponent(x), ...)`
- `ReactDOMRe.createDOMElementVariadic(x, ...)` -> `React.createElementVariadic(ReactDOM.stringToComponent(x), ...)`
- `ReactDOMRe.props` -> `ReactDOM.props`
- `ReactDOMRe.domProps` -> `ReactDOM.domProps`

PR #600, opened 2020-06-10
Add a way to create a DOM component without JSX

bloodyowl:


rickyvetter:
This makes a lot of sense. Thanks! Could you add docs for this if you get a chance?

peterpme:
Thank you!

PR #602, opened 2020-06-10
Update ReactDOM docs

bloodyowl:
Took the opportunity to:

- rename `ReactDOMRe` occurrences to `ReactDOM`
- remove `ReasonReact` bindings calls and replace them with modern `React` ones
- replace OCaml's `*_of_*` with Belt, so that it doesn't seem inconsistent in terms of casing
- added a mention of `React.string/int/float…` in the new docs (only appeared in the deprecated one)
- feature `ReactDOM.querySelector` instead of `renderToElementWith*` in docs


peterpme:
This is awesome, thank you!

PR #616, opened 2020-08-15
Fix A Simple Counter Example in Docs

seagalputra:
Hi,
I got error when trying execute example in useState Hook. I found that error caused by initialCount props is passing `int`, but it tried to render as `string` with `React.string()`.

Screen Shot 2020-08-15 at 18 56 04


peterpme:
Thanks!

PR #630, opened 2020-10-29
Update experimental bindings

sorenhoyer:
Config argument was removed from useTransition hook https://github.com/facebook/react/blob/master/packages/react/src/ReactHooks.js

createRoot, createBlockingRoot and useTransition got prefixed with "unstable_"

### DIFF
--- a/src/React.re
+++ b/src/React.re
@@ -525,10 +525,10 @@ module Uncurried = {
 type transitionConfig = {timeoutMs: int};
 
 [@bs.module "react"]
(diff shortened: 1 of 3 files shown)


tsnobip:
any update about when this could be merged? I'm interested too!

davesnx:
This changes aren't much relevant anymore (no unstable_*) and work has been done here to support React new versions: https://github.com/reasonml/reason-react/pull/630

PR #640, opened 2020-12-22
Fix #639: unused variable in website example

### LINKED ISSUE #639: Minimal example on ReasonReact website has an unused variable "name"
The ReasonReact website (https://reasonml.github.io/reason-react/) has the following minimal example on the front page:

```reasonml
[@react.component]
let make = (~name) =>
  <button>
    {React.string("Hello!")}
  </button>;
```

Here the prop `name` is not actually used. 

Maybe it's an idea to change this to render something like `{React.string("Hello" ++ name ++ "!")}`? 

😄

josdejong:
Fixes #639 .

I've changed `"Hello!"` to `"Hello " ++ name` and not `"Hello " ++ name ++ "!"` in order to keep the example as minimal as possible.

I'm not entirely sure whether I fixed the example in the right place, (in the source, and not in some generated output), please double check.

### DIFF
--- a/website/pages/en/index.js
+++ b/website/pages/en/index.js
@@ -32,7 +32,7 @@ const codeExampleSmallScreen =`${pre}reason
 [@react.component]
 let make = (~name) =>
   <button>
-    {React.string("Hello!")}
+    {React.string("Hello" ++ name)}
   </button>;
 ${pre}`;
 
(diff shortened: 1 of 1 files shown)


davesnx:
Thanks @josdejong

PR #641, opened 2020-12-31
Add drag event API binding

ersinakinci:
Add and use a new `ReactEvent.Drag` module for `onDrag`, `onDrop`, etc. The new module is like `ReactEvent.Mouse` but also [includes a binding for `dataTransfer`](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API#Interfaces).

### DIFF
--- a/src/ReactEvent.re
+++ b/src/ReactEvent.re
@@ -134,6 +134,36 @@ module Mouse = {
   [@bs.get] external shiftKey: t => bool = "shiftKey";
 };
 
+module Drag = {
+  type tag;
+  type t = synthetic(tag);
+  include MakeEventWithType({
+    type nonrec t = t;
+  });
+
+  // MouseEvent
+  [@bs.get] external altKey: t => bool = "altKey";
+  [@bs.get] external button: t => int = "button";
+  [@bs.get] external buttons: t => int = "buttons";
+  [@bs.get] external clientX: t => int = "clientX";
+  [@bs.get] external clientY: t => int = "clientY";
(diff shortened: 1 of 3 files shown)


src/ReactEvent.re
```diff
@@ -134,6 +134,36 @@ module Mouse = {
...
+  [@bs.get] external movementY: t => int = "movementY";
+  [@bs.get] external pageX: t => int = "pageX";
+  [@bs.get] external pageY: t => int = "pageY";
+  [@bs.get] [@bs.return nullable]
+  external relatedTarget: t => option(Js.t({..})) = "relatedTarget"; /* Should return Dom.eventTarget */
+  [@bs.get] external screenX: t => int = "screenX";
+  [@bs.get] external screenY: t => int = "screenY";
+  [@bs.get] external shiftKey: t => bool = "shiftKey";
+
+  [@bs.get] external dataTransfer: t => Dom.dataTransfer = "dataTransfer";
```
davesnx, line 164: `Dom.dataTransfer` is an opaque type, probably better to return Js.t({..}). Even though it's unsafe it's better to allow users to access by object accessor.

davesnx (approved):
There hasn't been any reason why Drag wasn't part of reason-react. I left a small comment about dataTransfer, which has been the case for other events that expose Js.t instead of the opaque type.

I know that I'm reviewing this PR super late, but that's what we got. If there's no interest from your part, I'm happy to push it.

PR #648, opened 2021-02-08
Fix typo

digression99:
Since context example includes JavaScript code example, fix typos for correction.


davesnx:
Thanks @digression99, better late than never :)

Issue #309, opened 2018-12-17
Support for aria-haspopup


laufeyrut:
The support for aria-haspopup is commented out:

https://github.com/reasonml/reason-react/blob/20c307f794298a514c7b23efbc02c3cdd21172d9/src/ReactDOMRe.re#L109

Is it possible to add support for this attribute? 🙏

mikaello:
This seems to be blocked by a feature missing in bucklescript, this is the comment rickyvetter made [when he merged the code](https://github.com/reasonml/reason-react/pull/228) you have linked:

> This adds all of the primitive aria types, **but saves the token ones for when bs.deriving abstract supports bs.string polymorphic variants**. Hopefully soon!

I think issue [[@bs.string] for bs.deriving abstract](https://github.com/BuckleScript/bucklescript/issues/2834) in the bucklescript repo is the one that needs to be resolved before `aria-haspopup` could be available.

jwhitley:
It seems unfortunate that the use of many aria attributes is blocked on a low-priority feature request, and one which (from comments on the linked issue) isn't considered approachable for new contributors.

Perhaps to unblock these aria attributes, instead of leaving them out of the API, they should all just be given type `string`?  Yes, that loses the ability to strongly type these attributes, and yes, it would mean a breaking change later, _but then people could use them to build accessible UI_.  /cc @rickyvetter

johnridesabike:
Unless the BuckleScript issue is getting resolved soon, then I agree that typing them as `string` seems like the best solution. In terms of type-safety, it a small compromise to make, and isn't any worse than regular ReactJS. But not being able to use them at all is a major problem for ReasonReact, IMO.

Issue #480, opened 2019-09-26
ariaChecked missing in the list of props for Dom elements


bsansouci:
Title says it all!

davesnx:
Better late than never. Merged in #697.

Issue #274, opened 2018-08-28
Add aria-current support


danben4:
Right now the support for `aria-current` is commented out: https://github.com/reasonml/reason-react/blob/6de6b7723aa4402922bd8717c2098868afe8570a/src/ReactDOMRe.re#L97

Is there any specific reason for this? Is it possible to add support for this attribute?

mikaello:
See my comment in https://github.com/reasonml/reason-react/issues/309#issuecomment-466050301 (the issue that is blocking `aria-*` is [[@bs.string] for bs.deriving abstract](https://github.com/BuckleScript/bucklescript/issues/2834) in the BuckleScript repo)

davesnx:
Better late than never. Merged in #697.

PR #697, opened 2022-12-16
Add all aria attributes

davesnx:
Closes https://github.com/reasonml/reason-react/issues/309, https://github.com/reasonml/reason-react/issues/274 and https://github.com/reasonml/reason-react/issues/480.

## Intro

This PR transforms all the data from MDN encoded in https://github.com/ml-in-barcelona/server-reason-react/blob/main/lib/domProps.ml#L232 and https://github.com/ml-in-barcelona/jsoo-react/blob/main/lib/dom_aria_attributes.ml into domProps.

Note: Didn't add `aria-dropeffect` since is deprecated (aria-grabbed is deprecated but it's present in the bindings).

## Solution

The approach of not adding most aria props was due to the weird types from the DOM API, and the impossibility of representing correctly those. For example, some of the attributes are `true | false | undefined` and those are encoded as strings. 

In this PR I treat everything that isn't straightforward an integer, string or number as a string. This way we can enable lower-case components to have aria attributes.

## Future

The plan to bring more type-safety on the lower-case components has always been important for me, currently on jsoo-react we detect which tag is used and rely on a list of valid attributes for each tag, giving an extra level of safety.

Another proposal might be to try to encode all strings that could be variants (or polyvariants), but let's not talk about future (and uncertain) plans.

## Problems

A downside of this PR is the increase of noise on error messages. Currently when there's a lower-case component with a prop that doesn't exist the error message is very ugly, with this PR it's going to be uglier (If I ever implement the same mechanism as jsoo-react, this problem disappears).

There are 3 props that are encoded a booleans while I have them as strings, due to the weirdness of the dom. I left them as bools, to avoid a breaking change.
```
ariaHidden, ariaExpanded, ariaSelected
```

### DIFF
--- a/src/ReactDOM.re
+++ b/src/ReactDOM.re
@@ -94,7 +94,7 @@ module Props = {
     [@bs.optional] [@bs.as "aria-errormessage"]
     ariaErrormessage: string, /* https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-errormessage */
     [@bs.optional] [@bs.as "aria-expanded"]
-    ariaExpanded: string, /* https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded */
+    ariaExpanded: bool, /* string */ /* https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded */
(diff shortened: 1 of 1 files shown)


rolandpeelen (approved):
This seems like a generated change - did not go over each individual one.

PR #708, opened 2023-03-20
Push old-new README

davesnx:
This PR tries to bring back minimum documentation to get started, install and use reason-react. Currently, the website is written with BuckleScript in mind and it will change to recommend using Melange or bs-platform@9 in the future.

Opening this PR as a draft early to discuss the "Warning" about ReScript users to be redirected to rescript as early as possible. cc @ryyppy @chenglou to confirm this is correct.

To see the rendered README: https://github.com/reasonml/reason-react/blob/Rollback-deprecation-warning-push-old-readme/README.md


davesnx:
There's has been some miss-conceptions lately on Discord. Merging this since I didn't get any feedback, happy to change the Warning on ReScript later.

PR #709, opened 2023-03-20
Add dune file

davesnx:
This PR enables the consumption of reason_react as melange dependency but it maintains the building with bs-platform and dune

### DIFF
--- a/src/dune
+++ b/src/dune
@@ -1,0 +1,5 @@
+(library
+ (name reason_react)
+ (wrapped false)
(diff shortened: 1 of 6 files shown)


src/dune
```diff
@@ -0,0 +1,5 @@
+(library
+ (name reason_react)
+ (wrapped false)
+ (flags :standard -bs-jsx 3)
```
anmonteiro, line 4: depending which melange version you install this could be changed to `(preprocessing (pps reactjs-jsx-ppx))`

PR #712, opened 2023-04-11
add react 17 as dev dependency

tatchi:


anmonteiro:
We could probably just replaced it? React 18 is on the roadmap but 17 could go in in the meantime.

tatchi:
> We could probably just replaced it? React 18 is on the roadmap but 17 could go in in the meantime.

But keeping only react 17 would mean that `reason-react` is not compatible with older versions, right? Maybe you got confused by my previous commit message that mentioned `devDependencies` instead of `peerDependencies` 🙈

davesnx:
Closing in favor of https://github.com/reasonml/reason-react/pull/761

jchavarri:
To add a bit more context, now that we have `opam-check-npm-deps` plugin, we can move the "peer dependencies" to the `depexts` field in the opam file.

PR #714, opened 2023-04-20
feat: react/jsx-runtime

anmonteiro:


src/jsx.re
```diff
@@ -0,0 +1,17 @@
+type element;
+type ref;
+
+[@bs.val] external null: element = "null";
+
+external float: float => element = "%identity";
+external int: int => element = "%identity";
+external string: string => element = "%identity";
+
+external array: array(element) => element = "%identity";
```
davesnx, line 10: You might need to `include Jsx` in React module, otherwise those functions won't be available under `React.*`

src/jsx.re
```diff
@@ -0,0 +1,17 @@
+type element;
+type ref;
```
davesnx, line 2: It's ref used for anything?

src/jsxDOM.re
```diff
@@ -0,0 +1,1059 @@
...
+
+/*** Internal: use JsxDOM directly. */
+
+/* type style = JsxDOMStyle.t; */
+type style = ReactDOMStyle.t;
+type domRef;
+
+/*
+   This list isn't exhaustive. We'll add more as we go.
+ */
```
davesnx, line 33: Does it make sense to specify in a comment the differences? I believe the type of children and ref are the only ones?

src/React.re
```diff
@@ -28,18 +28,30 @@ external createElementVariadic:
...
+
+/* [@bs.module "react/jsx-runtime"] */
+/* external jsxFragment: component('props) = "Fragment"; */
+[@bs.module "react/jsx-runtime"] external jsxFragment: 'a = "Fragment";
+/* external jsxFragment: component({. "children": element}) = "Fragment"; */
+/* [@bs.obj] */
+/* external makeProps: */
+/* (~children: element, ~key: 'key=?, unit) => {. "children": element}; */
+[@bs.module "react"]
+external make: component({. "children": element}) = "Fragment";
```
davesnx, line 53: I'm missing something here: this is under React and doesn't seem right, would it make more sense under a `Fragment` module (not really important the name since this is called from inside the ppx

davesnx (commented):
Left a few comments as if was ready to review, most of them are questions but I would avoid the change of the API under React.string, React.int, React.float, etc.

The rest looks good

src/ReactDOM.re
```diff
@@ -2117,3 +2117,25 @@ include Props;
...
+external jsxKeyed: (string, JsxDOM.domProps, string) => Jsx.element = "jsx";
+
+[@bs.module "react/jsx-runtime"]
+external jsx: (string, JsxDOM.domProps) => Jsx.element = "jsx";
+
+[@bs.module "react/jsx-runtime"]
+external jsxs: (string, JsxDOM.domProps) => Jsx.element = "jsxs";
+
+[@bs.module "react/jsx-runtime"]
+external jsxsKeyed: (string, JsxDOM.domProps, string) => Jsx.element = "jsxs";
```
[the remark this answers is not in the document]
anmonteiro: Which 2 sets? The signatures are different

src/React.re
anmonteiro: Why is this not supposed to be React.Fragment? Just like React.Suspense this should live here, no?

davesnx: Sure, but it's currently not under Fragment module.

src/ReactDOM.re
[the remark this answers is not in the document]
davesnx: Totally, but why we need the ReactDOM signatures to be different? do we use them on the ppx?

src/React.re
anmonteiro: oh I see what you're saying. I think this is an oversight, will look into it again.

src/ReactDOM.re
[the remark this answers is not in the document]
anmonteiro: Yes, the 3rd argument is `key`. We use it if the component passes a key prop.

anmonteiro:
Not really sure what these errors are about? 

```
FAIL _build/default/test/test/test/React__test.bs.js
  ● Test suite failed to run

    describe does not expect any arguments

       6 |
       7 | function describe(name, f) {
    >  8 |   Caml_external_polyfill.resolve("describe")(name, (function (param) {
         |                                             ^
       9 |           Curry._1(f, undefined);
      10 |         }));
      11 | }

      at Env.describe (node_modules/jest-jasmine2/build/jasmine/Env.js:402:19)
      at Object.describe (_build/default/test/test/test/jest/Jest.bs.js:8:45)
      at Object.<anonymous> (_build/default/test/test/test/React__test.bs.js:207:6)

FAIL _build/default/test/test/test/ReactRouter__test.bs.js
  ● Test suite failed to run

    describe does not expect any arguments

       6 |
       7 | function describe(name, f) {
    >  8 |   Caml_external_polyfill.resolve("describe")(name, (function (param) {
         |                                             ^
       9 |           Curry._1(f, undefined);
      10 |         }));
      11 | }

      at Env.describe (node_modules/jest-jasmine2/build/jasmine/Env.js:402:19)
      at Object.describe (_build/default/test/test/test/jest/Jest.bs.js:8:45)
      at Object.<anonymous> (_build/default/test/test/test/ReactRouter__test.bs.js:9:6)
```


will investigate.

davesnx:
There seems to be some issue with bindings, when building 

```
File "test/React__test.re", lines 563-567, characters 4-18:
563 | ....expect(
564 |       container
565 |       ->DOM.findBySelectorAndPartialTextContent("div", "10")
566 |       ->Option.isSome,
567 |     )->toBe(true);
Warning 106 [bucklescript-unimplemented-primitive]: Unimplemented primitive used:toBe
```

anmonteiro:
Ah, makes sense, was missing the ppx https://github.com/reasonml/reason-react/pull/714/commits/0da7411d56cfe4a0625f3bcaadb24135c9dccc26

PR #715, opened 2023-04-24
Deprecate legacy folder modules

davesnx:
Kept all compatibility to the ppx so no changes need to be made to ppx.ml neither https://github.com/melange-re/melange/tree/main/reactjs-jsx-ppx


anmonteiro (approved):
not really a "deprecation" :P

PR #722, opened 2023-05-31
chore: port react js ppx from the melange repo

anmonteiro:
ports a few changes we had over there


ppx/src/reactjs_jsx_ppx.ml
```diff
@@ -500,7 +500,7 @@ let jsxMapper =
     Exp.apply
```
anmonteiro, line 500: the file rename matches `(library (name ...))` now, that's why I made that change.

PR #728, opened 2023-06-01
chore: comment out test

anmonteiro:
I'm not sure what exactly to do about this, but it looks like a) writing this melange test and b) having reactjs-ppx in this repository will require dune 3.9 because of https://github.com/ocaml/dune/pull/7849.

My proposal is to comment it out until we can require `(dune lang 3.9)` after the next dune release. In the meantime we can uncomment it to run the tests for local development.

What do you think, @davesnx?

### DIFF
--- a/.github/workflows/opam.yml
+++ b/.github/workflows/opam.yml
@@ -65,8 +65,8 @@ jobs:
       - name: Build
         run: make build
 
-      - name: Test
-        run: make test
+      # - name: Test
+        # run: make test
 
       - name: Save cache when not Windows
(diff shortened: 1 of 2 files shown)


davesnx:
Happy to keep the tests skipped until this fix lands 👍🏼

jchavarri:
> My proposal is to comment it out until we can require (dune lang 3.9) after the next dune release.

There might be a point release 3.8.1 soon as there was some regression related to Coq, maybe the fix in 7849 will be included in that release.

PR #732, opened 2023-06-12
Rename reactjs-jsx-ppx to reason-react-ppx

davesnx:
Renamed the ppx to make it more clear what it is. I didn't update the file-name due to some further PRs touching this and not causing conflicts.

@anmonteiro updated flake.nix but I'm not able to test it locally, could you make sure it's alright?


anmonteiro:
Should this be called `ppx_react_jsx`? I think it's more common to prefix ppx library names with `ppx_*` rather than suffix?

davesnx:
Happy to follow the convention. I remember discussing the name previously to something like `reason-react.ppx` but we wanted a different package. Does `ppx_reason_react` sound good?

anmonteiro:
Right, we wanted a different package so that we'd be able to install the ppx separately from the library.


I think `ppx_reason_react` makes sense, since the PPX is specific to this library.

PR #733, opened 2023-06-12
Post v0.11 release cleanup

davesnx:
- Enable tests back
- Update package.json version

### DIFF
--- a/.github/workflows/opam.yml
+++ b/.github/workflows/opam.yml
@@ -47,7 +47,7 @@ jobs:
           key: opam-${{ matrix.os }}-${{ matrix.ocaml-compiler }}-${{ hashFiles('**.opam') }}
 
       - name: Use OCaml ${{ matrix.ocaml-compiler }}
-        uses: ocaml/setup-ocaml@v2
+        uses: ocaml/setup-ocaml@v2.0.15
         with:
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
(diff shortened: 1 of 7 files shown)


.github/workflows/opam.yml
```diff
@@ -49,7 +47,7 @@ jobs:
           key: opam-${{ matrix.os }}-${{ matrix.ocaml-compiler }}-${{ hashFiles('**.opam') }}
 
       - name: Use OCaml ${{ matrix.ocaml-compiler }}
-        uses: ocaml/setup-ocaml@v2
+        uses: ocaml/setup-ocaml@v2.0.15
```
anmonteiro, line 50: this should be the same as before

test/dune
```diff
@@ -1,9 +1,9 @@
...
-; (target test)
-; (package reason-react)
-; (module_systems
-; (commonjs bs.js))
-; (libraries reason_react jest)
-; (preprocess
-; (pps melange.ppx reactjs-jsx-ppx)))
+(melange.emit
+ (alias runtest)
+ (target melange)
```
anmonteiro, line 3: a bit confusing to call everything `melange`. maybe `test-output`?

.github/workflows/opam.yml
davesnx: I was trying to see if updating the action will fix the windows error :(

PR #744, opened 2023-07-11
Fix installation and tests

davesnx:
Install Reason v3.9.0 on make install and run snapshots with it.


davesnx:
We don't but this just installs the same version on local than on CI. The requirement is still 3.6.0

dune-project
```diff
@@ -1,4 +1,4 @@
-(lang dune 3.8)
+(lang dune 3.9)
```
anmonteiro, line 1: This change is clearly requiring 3.9

davesnx: Oh, thought you talked about Reason, not dune.

anmonteiro: But melange only requires dune 3.8

davesnx: The current state of this repo is very broken, since it requires melange master and Reason 3.9.0 for the snapshot testing + the reason-react-ppx from the repo

So there should be a few things to un-tangle before we can even publish newer versions

PR #748, opened 2023-07-13
Better editor and merlin integration by improving ppx locations

### LINKED ISSUE #429: merlin and RLS broken for JSX3 components
https://youtu.be/Hf6hEqHqpG4 shows the issues for ReasonReact JSX3 + GraphQL, but this is also the case even for simple components:

```reason
open React;
open! ReasonReact;

type state = {
  count: int,
  show: bool,
};

type action =
  | Click(int)
  | Toggle;

let initialState = {count: 0, show: true};

[@react.component]
let make = (~greeting) => {
  let (state, dispatch) =
    useReducer(
      (state, action) =>
        switch (action) {
        | _ => state
        | Click(points) => {...state, count: state.count + points}
        | Toggle => {...state, show: !state.show}
        },
      initialState,
    );

  let message =
    "You've clicked this " ++ string_of_int(state.count) ++ " times(s)";

  <div>
    <button onClick={_ => dispatch(Click(10))}> {string(message)} </button>
    <button onClick={_ => dispatch(Toggle)}>
      {string("Toggle greeting")}
    </button>
    {state.show ? string(greeting) : null}
  </div>;
};
```

The issue for this smaller component show up in weird places. Asking for the type under lots of locations works, but asking for the type of `points` under  `| Click(points) =>` expands the selection to the whole function (everything after `let make = `. Same thing for checking `dispatch` in `let (state, dispatch)`, though `state` works fine.

Similarly for `{state.show ? string(greeting) : null}`, these tokens work: `state`, `string`, `null`
and these expand the selection to the full component `show` (in `state.show`), `greeting`

Screenshot 2019-05-17 18 51 01
Screenshot 2019-05-17 18 50 43

jchavarri:
Fixes #429.

The fix is kind of similar to the one used [in jsoo-react](https://github.com/ml-in-barcelona/jsoo-react/pull/134), but adapted to the current state of this ppx.

It is based on these ideas:

- Avoid using `Location.none`, especially on intermediate ast nodes. Using `Location.none` completely disables merlin ability to find back an ast node for a given location
- Try to build locations in a consistent way with the ast tree. for what "consistent" means, check [this explanation](https://github.com/ocaml/ocaml/pull/8987) by trefis
- Hide subparts of the ast with `merlin.hide` (when we want to hide whole subtree) or `loc_ghost` (for just one node)
- Use `-dparsetree` in tests to both check which locations are set to none or inconsistent, and see how changes in ppx affect the result

~The PR adds some tests using `-dparsetree` to see how changes in the ppx affect locations in real code. Unfortunately I am not able to set up a test that involves merlin as well, so what I do for testing, besides looking at the output in the newly added `ppx/test/output_locations.expected`, is:~
~- Change `reactjs_jsx_ppx.ml`~
~- `make build`~
~- Hover over elements in `test/React__test.re`~

~This is a quite fast workflow.~ Edit: tests using merlin directly are being added in #749.

Here are some gifs to show comparison between `main` and this branch.

### Case 1: Component in `React__test.re`

| `main`  | This branch |
| ------------- | ------------- |
| case-1-ppx-locs-before | case-1-ppx-locs-after |

### Case 2: The component reported in original issue #429

| `main`  | This branch |
| ------------- | ------------- |
| case-2-ppx-locs-before | case-2-ppx-locs-after |

### DIFF
--- a/ppx/src/reactjs_jsx_ppx.ml
+++ b/ppx/src/reactjs_jsx_ppx.ml
@@ -762,18 +762,11 @@ let jsxMapper =
     (* let component = ... *)
     | { pstr_loc; pstr_desc = Pstr_value (recFlag, valueBindings) } ->
         let fileName = filenameFromLoc ~ctxt pstr_loc in
-        let emptyLoc = Location.in_file fileName in
+        let gloc = { pstr_loc with loc_ghost = true } in
         let mapBinding binding =
           if hasAttrOnBinding binding then
             let bindingLoc = binding.pvb_loc in
             let bindingPatLoc = binding.pvb_pat.ppat_loc in
-            let binding =
-              {
-                binding with
-                pvb_pat = { binding.pvb_pat with ppat_loc = emptyLoc };
-                pvb_loc = emptyLoc;
-              }
-            in
             let fnName = getFnName binding in
             let internalFnName = fnName ^ "$Internal" in
             let fullModuleName =
@@ -827,8 +820,7 @@ let jsxMapper =
               let unerasableIgnoreExp exp =
                 {
                   exp with
-                  pexp_attributes =
-                    unerasableIgnore emptyLoc :: exp.pexp_attributes;
+                  pexp_attributes = unerasableIgnore gloc :: exp.pexp_attributes;
                 }
               in
               (* TODO: there is a long-tail of unsupported features inside of
@@ -947,7 +939,7 @@ let jsxMapper =
               match reactComponentAttribute with
               | Some { attr_name = loc; attr_payload = payload; _ } ->
                   (loc.loc, Some payload)
-              | None -> (emptyLoc, None)
+              | None -> (gloc, None)
             in
             let props = getPropsAttr payload in
             (* do stuff here! *)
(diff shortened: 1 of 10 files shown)


jchavarri:
There was a comment by Ricky about some tradeoff between optimizing for good errors or for merlin integration [in the original issue](https://github.com/reasonml/reason-react/issues/429#issuecomment-567561769).

Unfortunately no examples of this tradeoff were provided, I have tested with unused props, go to definition in props, type errors in props, etc. and all the tests I did were successful, in all cases both Merlin and the compiler can figure things out fine.

ppx/src/reactjs_jsx_ppx.ml
```diff
@@ -122,6 +122,15 @@ let unerasableIgnore loc =
     attr_loc = loc;
   }
 
+(* [merlinHide] tells merlin to not look at a node, or at any of its
+   descendants. *)
+let merlinHide =
+  {
```
anmonteiro, line 128: this is only ever used as a list with a single element. we could allocate that list just once here.

jchavarri: good point, fixed in https://github.com/reasonml/reason-react/pull/748/commits/ea0292763721fc82e95d16562bf4325af8c71bbb.

PR #749, opened 2023-07-13
ppx: add merlin tests

jchavarri:
Related to #748.

This PR adds a test suite leveraging Dune cram tests where `ocamlmerlin` is called directly to gather type information by sending some locations in reason-react components.

This approach is much less noisy than the one with `-dparsetree` originally used in #748. Ideally this PR should be merged before that one, so we can evaluate the changes after the location improvements over there.

### DIFF
--- a/ppx/test/merlin/issue-429.t/component.re
+++ b/ppx/test/merlin/issue-429.t/component.re
@@ -1,5 +1,4 @@
 open React;
-open! ReasonReact;
 
 type state = {
   count: int,
--- a/ppx/test/merlin/issue-429.t/component.re
+++ b/ppx/test/merlin/issue-429.t/component.re
@@ -1,0 +1,37 @@
+open React;
+open! ReasonReact;
+
+type state = {
+  count: int,
+  show: bool,
+};
+
+type action =
+  | Click(int)
+  | Toggle;
+
+let initialState = {count: 0, show: true};
+
+[@react.component]
+let make = (~greeting) => {
+  let (state, dispatch) =
+    useReducer(
+      (state, action) =>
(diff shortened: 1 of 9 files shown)


ppx/test/merlin/issue-429.t/component.re
```diff
@@ -0,0 +1,37 @@
+open React;
+open! ReasonReact;
```
anmonteiro, line 2: why do we need this `open`?

ppx/test/merlin/issue-429.t/run.t
```diff
@@ -0,0 +1,173 @@
+Test some locations in reason-react components
+
+  $ cat >dune-project <<EOF
+  > (lang dune 3.9)
```
anmonteiro, line 4: should be consistent with what we require in the opam file

```suggestion
  > (lang dune 3.8)
```

ppx/test/merlin/issue-429.t/component.re
jchavarri: i just copied the file from the original report in #429.

anmonteiro: I think `ReasonReact` might be going away soon. I don't think we need that module at all.

anmonteiro (approved):
nice. seems like a sustainable way to test

ppx/test/merlin/issue-429.t/component.re
jchavarri: Removed in https://github.com/reasonml/reason-react/pull/749/commits/1463abba06c7126610db86c531768b74ddb85b4a.

PR #750, opened 2023-07-14
test: add react.fragment with key

jchavarri:
It seems there's been a recent regression that involves `React.Fragment` and `key` attribute. It was possible to use `option(string)` before while now the code does not type check.

### DIFF
--- a/test/React__test.re
+++ b/test/React__test.re
@@ -589,6 +589,27 @@ describe("React", () => {
     expect(value.contents)->toEqual("My value");
   });
 
+  test("React.Fragment with key", () => {
+    let container = getContainer(container);
+    let title = Some("foo");
+
+    act(() => {
+      ReactDOM.render(
+        <React.Fragment key=?title>
+          <div> "Child"->React.string </div>
+        </React.Fragment>,
+        container,
+      )
+    });
+
+    expect(
+      container
+      ->DOM.findBySelectorAndPartialTextContent("div", "Child")
+      ->Option.isSome,
+    )
+    ->toBe(true);
+  });
+  
   /* test("ErrorBoundary", () => {
     let container = getContainer(container);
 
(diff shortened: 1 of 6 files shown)


test/React__test.re
```diff
@@ -589,6 +589,27 @@ describe("React", () => {
     expect(value.contents)->toEqual("My value");
   });
 
+  test("React.Fragment with key", () => {
+    let container = getContainer(container);
+    let title = Some("foo");
+
+    act(() => {
+      ReactDOM.render(
+        <React.Fragment key=title>
```
anmonteiro, line 598: there's definitely a bug in reason-react, but the test is also wrong. this should be `key=?title`

anmonteiro: `Fragment` and some other special components seem very lax in what they accept. we need to make it only take string (option)s. I'm pushing a fix momentarily.

jchavarri: > but the test is also wrong

 Yeah the test looks wrong, but I wanted to reflect the code as it was working before.

jchavarri: updated.

jchavarri:
Found some interesting change when testing this branch. 

The following code would compile before without issues:

```reason
let resultRow = twitterAuthor =>
  <tr key={twitterAuthor.EndpointT.screen_name}>
    <td className="DetailsWhoTweeted-col--avatar">
      <div className=Css.avatarContainer>
        <CeAvatarIcon
          src={Some(String.replace(twitterAuthor.profile_image_url->Wrap.Url.show, "http://", "https://"))}
        />
      </div>
    </td>
  </tr>
```

With this branch (or maybe, after the changes in https://github.com/reasonml/reason-react/pull/714? not sure), it triggers this error:

```
File "CeDetailsWhoTweeted.re", line 154, characters 51-68:
154 |             src={Some(String.replace(twitterAuthor.profile_image_url->Wrap.Url.show, "http://", "https://"))}
                                                         ^^^^^^^^^^^^^^^^^
Error: Unbound record field profile_image_url
```

jchavarri:
@anmonteiro I updated the branch with latest `main`. On my side everything looks good, the issue reported [above](https://github.com/reasonml/reason-react/pull/750#issuecomment-1639756688) doesn't seem like a big deal. 

Feel free to merge whenever you think is ready (I can't approve as the creator of the branch).

PR #752, opened 2023-07-18
ppx: add test for inference regression in keys

jchavarri:
This test shows some regression where the type checker is not able to infer the type of a prop for a value which type has been inferred before in a `key` prop.

The compilation fails with:

```
File "test/React__test.re", line 605, characters 26-34:
605 |           <img src=author.imageUrl />
                                ^^^^^^^^
Error: Unbound record field imageUrl
make: *** [Makefile:12: build] Error 1
```

When before, it would infer that `author` has type `Author.t`, as it is namespaced in line 603:

https://github.com/reasonml/reason-react/blob/c9acf14b6cbf2dbd27869cc5d41e255f5b7961c3/test/React__test.re#L603

I bisected the test, and this used to work before #714 was merged (tested with commit 2b8819bb6b59932116af24d0571729635e0e12b6 and it builds fine)


anmonteiro:
Check the difference in generated code.

Before:

```ocaml
let render author =
  ReactDOMRe.createDOMElementVariadic "tr"
    ~props:(ReactDOMRe.domProps ~key:author.Author.name ())
    [|
      ReactDOMRe.createDOMElementVariadic "td"
        [|
          ReactDOMRe.createDOMElementVariadic "img"
            ~props:(ReactDOMRe.domProps ~src:author.imageUrl ())
            [||];
        |];
    |]
```

After:

```ocaml
let render author =
  ReactDOM.jsxKeyed "tr"
    ((ReactDOM.domProps [@merlin.hide])
       ~children:
         (ReactDOM.jsx "td"
            ((ReactDOM.domProps [@merlin.hide])
               ~children:
                 (ReactDOM.jsx "img"
                    ((ReactDOM.domProps [@merlin.hide]) ~src:author.imageUrl ()))
               ()))
       ())
    ~key:author.Author.name ()
```

I tried this too, but it doesn't fix it:

```ocaml
let render author =
  ReactDOM.jsxKeyed ~key:author.Author.name "tr"
    ((ReactDOM.domProps [@merlin.hide])
       ~children:
         (ReactDOM.jsx "td"
            ((ReactDOM.domProps [@merlin.hide])
               ~children:
                 (ReactDOM.jsx "img"
                    ((ReactDOM.domProps [@merlin.hide]) ~src:author.imageUrl ()))
               ()))
       ())
    ()
```

jchavarri:
Thanks for sharing the ppx expanded.

With React jsx-runtime API, `key` is the last argument in this function:

https://github.com/reasonml/reason-react/blob/c9acf14b6cbf2dbd27869cc5d41e255f5b7961c3/src/ReactDOM.re#L2139

(side note: should it be a `key` labeled arg?)

So, because what matters for inference is the ordering of arguments at the definition, there is not much we can do. Here is a simpler example, note how the call to `foo` succeeds while `foo2` fails:

```
let foo = (~bar, ~baz) => bar ++ baz;
let foo2 = (~baz, ~bar) => bar ++ baz;
let _ = author => foo(~bar=author.Author.name, ~baz=author.imageUrl);
let _ = author => foo2(~bar=author.Author.name, ~baz=author.imageUrl);
                                                            ^^^^^^^^
Error: Unbound record field imageUrl
```

Funny enough (but understandably) this works fine:

```reason
let render = author =>
  <tr key={author.name}>
    <td>
      <img src=author.Author.imageUrl />
    </td>
  </tr>;
```

The only solution I can think of is to redefine the args order by doing something like:

```reason
[@bs.module "react/jsx-runtime"]
external jsxKeyed: (string, domProps, string) => React.element = "jsx";
let jsxKeyed(element, key, props) = jsxKeyed(element, props, key);
```

But I personally don't think it's worth it. Wdyt?

anmonteiro:
Ah you’re right, of course. I think this creates an extra function call for every JSX call though, which I don’t think is desirable.

anmonteiro:
@jchavarri I pushed a different version of your fix that generates:

```ocaml
let render author =
  let reason_react_ppx_key_arg___x = author.Author.name in
  ReactDOM.jsxKeyed ~key:reason_react_ppx_key_arg___x "tr"
    (((ReactDOM.domProps)[@merlin.hide ])
       ~children:(ReactDOM.jsx "td"
                    (((ReactDOM.domProps)[@merlin.hide ])
                       ~children:(ReactDOM.jsx "img"
                                    (((ReactDOM.domProps)[@merlin.hide ])
                                       ~src:(author.imageUrl) ())) ())) ())
    ()
```

we introduce a let binding that matches the expected flow of type inference. What do you think?

ppx/src/reactjs_jsx_ppx.ml
```diff
@@ -457,7 +457,8 @@ let makeExternalDecl fnName loc namedArgListWithKeyAndRef namedTypeList =
 
 (* TODO: some line number might still be wrong *)
 let jsxMapper =
-  let unit = Exp.construct { txt = Lident "()"; loc = Location.none } None in
+  let unit = Exp.construct { txt = Lident "()"; loc = Location.none } None
+  and key_var_txt = "reason_react_ppx_key_arg___x" in
```
jchavarri, line 461: Why the suffix `___x`?

ppx/src/reactjs_jsx_ppx.ml
```diff
@@ -499,18 +500,30 @@ let jsxMapper =
...
-      | Some (label, key) ->
-          [ (label, mapper#expression ctxt key); (nolabel, unit) ]
-      | None -> []
-    in
-    Exp.apply ~loc ~attrs jsxExpr
-      ([ (nolabel, Exp.ident ~loc { txt = ident; loc }); (nolabel, props) ]
-      @ key_args)
+    match key with
+    | Some (label, key) ->
+        Exp.let_ ~loc Nonrecursive
```
jchavarri, line 510: I would add a comment on why we need this `let` expression. Otherwise we (or at least I) will forget in just a few weeks 😅

ppx/src/reactjs_jsx_ppx.ml
```diff
@@ -522,33 +535,42 @@ let jsxMapper =
...
+         | None -> nonChildrenProps)
+        |> List.map (fun (label, expression) ->
+               (label, mapper#expression ctxt expression)))
     in
-    Exp.apply ~loc:parentExpLoc ~attrs jsxExpr args
+    let component = (nolabel, componentNameExpr)
+    and props = (nolabel, propsCall) in
+    match key with
+    | Some (label, key) ->
+        Exp.let_ ~loc Nonrecursive
```
jchavarri, line 557: Same (about the comment)

ppx/src/reactjs_jsx_ppx.ml
```diff
@@ -522,33 +535,42 @@ let jsxMapper =
...
+          [
+            {
+              pvb_pat = Pat.var { txt = key_var_txt; loc };
+              pvb_expr = mapper#expression ctxt key;
+              pvb_attributes = [];
+              pvb_loc = loc;
+            };
+          ]
+          (Exp.apply ~loc:parentExpLoc ~attrs jsxExpr
```
jchavarri, line 566: This kind of breaks the rules for "well-formed" locations, because the `apply` will have a larger range (`parentExpLoc`) than the parent `let` expression (which has `loc`). I am afraid that our for now limited tests are not catching a potential regression here.

I think we should at least rename `loc` to `callerLoc` or something else, to prevent this kind of issues in the future.

jchavarri (commented):
Your approach looks good as well! I left a comment about locations that I think should be tackled before merging.

In general, all other things equal, I tend to prefer solving issues elsewhere than into a ppx, just because it has so much inherent complexity... but it is not a strong opinion, happy to go with this solution.

Do you know which one would be more optimal in terms of resulting size of the generated JavaScript code? I understand the extra `let` in the bindings is just added once in the bindings, but the ppx will add an extra `let` for every usage of `key`?

ppx/src/reactjs_jsx_ppx.ml
anmonteiro: to minimize the chance of shadowing.

ppx/src/reactjs_jsx_ppx.ml
davesnx: The rename make sense, but I don't understand the issue here. I belive the location is pointed to the right place no?

```
(li ~key:e.path [@JSX]) (* <-- parentExpLoc *)

let reason_react_ppx_key_arg___x = e.path in
ReactDOM.jsxKeyed ~key:reason_react_ppx_key_arg___x "li" (* <-- parentExpLoc *)
```

davesnx: After pressing "Send", I realised that the `let reason_react_ppx_key_arg___x` use the location of the identifier, and this is the reason why when you hover a lowercase component it says "string"

jchavarri: In my experience, it doesn't matter that much which is the right location for each node. As long as they are well formed (parents have larger scope than children), things kind of work. One might need to hide some nodes to prioritize others, but the fundamental first step is that the ranges are well formed. When that fails, merlin will fail in most cases.

Issue #742, opened 2023-07-10
Use createRoot from React 18 and render from react-dom/client


psb:
Using
```reason
| Some(root) => ReactDOM.render(<App />, root) 
```
gives the warning:
```sh
ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17
```

Using
```reason 
| Some(root) => {
        let app = ReactDOM.Experimental.createRoot(root);
        ReactDOM.Experimental.render(app, <App />);
      }
```
gives the warning:
```sh
Warning: You are importing createRoot from "react-dom" which is not supported. You should instead import it from "react-dom/client".
```

Remove `Experimental` module from `src/ReactDOM.re` and use `createRoot` from React 18 and `render` from `react-dom/client`.

PR #756, opened 2023-07-28
Add React 18 bindings 

davesnx:
Reference https://react.dev/blog/2022/03/29/react-v18

Fixes https://github.com/reasonml/reason-react/issues/742

- [x] Add hydrateRoot/renderRoot (move away from Experimental)
- [x] React.useTransition
- [x] React.useDeferredValue
- [x] React.useId
- [x] React.useSyncExternalStore
- [x] React.useInsertionEffect
- [x] Remove `<SupenseList />`
- [x] ReactDOM.Server.renderToPipeableStream (aliased `ReactDOM.Server` to `ReactDOMServer`)
- [ ] ~`React.lazy()`~
- [x] Add warnings for ReactDOM.* methods
- [x] Refactor tests to avoid using ReactDOM.render
- [x] Added `React.Experimental.use`


src/ReactDOMServer.re
```diff
@@ -4,3 +4,24 @@ external renderToString: React.element => string = "renderToString";
 [@bs.module "react-dom/server"]
 external renderToStaticMarkup: React.element => string =
   "renderToStaticMarkup";
+
+type error = {.};
+
+[@bs.module "react-dom/server"]
+external renderToPipeableStream:
```
anmonteiro, line 11: this seems wrong. According to the TypeScript type definitions it should be a function of 2 arguments: a ReactNode and an options object:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f1c7be0bf694f55a4cb8b34bd1134e3d23cf21e7/types/react-dom/server.d.ts#L27

davesnx: Totally, I had copied my implementation from server-reason-react and forgot to bind it properly

src/React.re
```diff
@@ -528,16 +555,16 @@ module Uncurried = {
...
   "useTransition";
 
 [@bs.set]
 external setDisplayName: (component('props), string) => unit = "displayName";
 
 [@bs.get] [@bs.return nullable]
 external displayName: component('props) => option(string) = "displayName";
+
+[@bs.module "react"]
+external useDebugValue: (~format: 'value => string, 'value) => unit =
```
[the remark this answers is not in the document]
davesnx: Woops, I dislike the unit when the labelled arg is last. `useDebugValue: string` and `useDebugValueFormat: 'value, 'value => string` sounds good here?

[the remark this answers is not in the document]
davesnx: or just a wrapper with an idendity function by default
```
[@bs.module "react"]
external useDebugValue: ('value, ~format: 'value => string=?, unit) => unit =
  "useDebugValue";

let useDebugValue = (value, ~format=(a => a)) =>
  useDebugValue(value, ~format, ());
```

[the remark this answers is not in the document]
anmonteiro: what's the issue with the unit argument? It looks like `format` is optional so we can't do your first suggestion.

IMO the `unit` argument is idiomatic in OCaml for functions that have optional labeled arguments.

src/ReactDOMServer.re
```diff
@@ -4,3 +4,68 @@ external renderToString: React.element => string = "renderToString";
...
+  onShellError: option(error => unit),
+  [@bs.optional]
+  progressiveChunkSize: option(int),
+};
+
+[@bs.module "react-dom/server"]
+external renderToPipeableStream: (React.element, options) => string =
+  "renderToPipeableStream";
+
+let renderToPipeableStream =
```
anmonteiro, line 40: shouldn't most of these be optionally labeled arguments?

src/React.re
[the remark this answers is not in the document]
davesnx: There's no issue, just a preference and trying to avoid it.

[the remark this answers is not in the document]
anmonteiro: It's still more idiomatic in OCaml to have an optionally labelled argument despite of your personal preference, though.

I'd prefer a single function rather than 2. In Melange, the unit argument is ignored in the FFI.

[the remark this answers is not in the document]
davesnx: Yes, totally pushed the version with the unit

src/ReactDOMServer.re
davesnx: Totally, fixed in [`8347855` (#756)](https://github.com/reasonml/reason-react/pull/756/commits/834785584f01af3a95fb19fcf4f2dff6717c5127)

anmonteiro:
Do you want to get this in for the upcoming release?

anmonteiro:
@davesnx I took the liberty to rebase this based on the latest changes in main

src/React.rei
```diff
@@ -519,20 +560,25 @@ module Uncurried: {
...
-
 [@mel.module "react"]
-external useTransition:
-  (~config: transitionConfig=?, unit) =>
-  (callback(callback(unit, unit), unit), bool) =
+external useTransition: unit => (bool, callback(callback(unit, unit), unit)) =
   "useTransition";
 
+module Experimental: {
+  /* This module is used to bind to APIs for future versions of React. There are no garanties of backwards compatibiltiy or stability. */
```
anmonteiro, line 568:
```suggestion
  /* This module is used to bind to APIs for future versions of React. There is no guarantee of backwards compatibility or stability. */
```

src/React.rei
```diff
@@ -519,20 +560,25 @@ module Uncurried: {
...
-type transitionConfig = {timeoutMs: int};
-
 [@mel.module "react"]
-external useTransition:
-  (~config: transitionConfig=?, unit) =>
-  (callback(callback(unit, unit), unit), bool) =
+external useTransition: unit => (bool, callback(callback(unit, unit), unit)) =
   "useTransition";
 
+module Experimental: {
```
anmonteiro, line 567: should we also include SuspenseList in this module?

src/React.rei
```diff
@@ -1511,27 +1557,84 @@ module DOM: {
...
-    [@mel.send] external render: (root, element) => unit = "render";
+    [@mel.module "react-dom/client"]
+    external createRoot: Dom.element => root = "createRoot";
+
+    [@mel.module "react-dom/client"]
+    external hydrateRoot: (Dom.element, element) => root = "hydrateRoot";
   };
 
+  [@deprecated
+    "Please use React.DOM.Client.render instead. It will be removed in the next release. Disable this warning with `[@alert \"-deprecated\"]` before React.DOM.render."
```
anmonteiro, line 1630:
```suggestion
    "This function will be removed in the next release. Use React.DOM.Client.render instead."
```

src/React.rei
```diff
@@ -1511,27 +1557,84 @@ module DOM: {
...
+  ]
+  [@mel.module "react-dom"]
+  external render: (element, Dom.element) => unit = "render";
+
+  [@deprecated
+    "Please use React.DOM.Client.hydrateRoot instead. It will be removed in the next release. Disable this warning with `[@alert \"-deprecated\"]` before React.DOM.hydrate."
```
anmonteiro, line 1636:
```suggestion
    "This function will be removed in the next release. Use React.DOM.Client.hydrateRoot instead."
```

src/hooks.re
```diff
@@ -335,10 +406,16 @@ module Uncurried = {
...
-  (callback(callback(unit, unit), unit), bool) =
+external useTransition: unit => (bool, callback(callback(unit, unit), unit)) =
   "useTransition";
+
+[@mel.module "react"]
+external useDebugValue: ('value, ~format: 'value => string=?, unit) => unit =
+  "useDebugValue";
+
+module Experimental = {
+  /* This module is used to bind to APIs for future versions of React. There are no garanties of backwards compatibiltiy or stability. */
```
anmonteiro, line 418:
```suggestion
  /* This module is used to bind to APIs for future versions of React. There is no guarantee of backwards compatibility or stability. */
```

anmonteiro (approved):

src/React.rei
davesnx: SuspenseList got unstable by React in https://github.com/facebook/react/pull/27061 and https://github.com/facebook/react/pull/27056#discussion_r1253282784

I'm fine keeping it but it's very useless

davesnx:
We would merge this and release under the 0.14 version!

PR #760, opened 2023-08-24
add npm depexts

jchavarri:
I am not sure if 0.11 is also compatible with react 17 🤔 but it seems safe to limit the range to just 16?


davesnx (approved):

jchavarri:
@davesnx I remove `peerDependencies` in last commit, as that is effectively what we do with `depexts` now.

PR #765, opened 2023-09-03
Migrate snapshot testing to cram tests

davesnx:
As usual, prefer cram tests over dune rules + diff


davesnx:
Fucked up history, reopening clean in: https://github.com/reasonml/reason-react/pull/767

PR #767, opened 2023-09-04
Migrate snapshot to cram

davesnx:


ppx/dune
```diff
@@ -0,0 +1,14 @@
...
+ (public_name reason-react-ppx)
+ (flags :standard -w -9)
+ (kind ppx_rewriter)
+ (libraries ppxlib))
+
+(executable
+ (name standalone)
+ (modules standalone)
+ (package reason-react-ppx)
+ (public_name reason-react-ppx.standalone)
```
jchavarri, line 13: Is there any reason to make this public? I think the ppx will always be consumed as a library.

ppx/test/functor.t/run.t
```diff
@@ -0,0 +1,20 @@
+  $ ../ppx.sh input.re
```
jchavarri, line 1: Why sometimes `ppx_with_refmt` is used and some other times just `ppx`?

jchavarri (approved):
nice!

ppx/dune
anmonteiro: It can be useful to run the ppx transformation as an executable. I'm in favor of this one.

I'd name it to a better name though. I think it can even be called `reason-react-ppx`. it'll be installed in `bin/reason-react-ppx`.

davesnx: I didn't know was possible the same name as the lib

ppx/test/issue-429.t/component.re
jchavarri, line 1: I just noticed one downside of this rearrangement is that now merlin tests are mixed up with the other ppx tests. I would open a PR to put them back together because the tests are quite different and it was nice to have them grouped together.

PR #768, opened 2023-09-04
Abstract bindings on ppx

davesnx:
Previously to remove the legacy APIs I want to make sure we don't rely on those on the ppx. I'm doing this small abstraction to have an easy way to know which bindings we use from the ppx and improving a big the quality of this beast.

This refactor should be harmless. I have tested the location test and seems ok.

- Abstract bindings into Binding
- Remove Ast_helper usages


ppx/src/reactjs_jsx_ppx.ml
```diff
@@ -1231,7 +1288,7 @@ let jsxMapper =
               parentExpLoc attrs callExpression callArguments
         (* div(~prop1=foo, ~prop2=bar, ~children=[bla], ()) *)
         (* turn that into
-           ReactDOMRe.createElement(~props=ReactDOMRe.props(~props1=foo,
+           ReactDOM.createElement("div", ~props=ReactDOM.domprops(~props1=foo,
```
jchavarri, line 1291:
```suggestion
           ReactDOM.createElement("div", ~props=ReactDOM.domProps(~props1=foo,
```

ppx/test/output_locations.expected
```diff
@@ -750,10 +752,10 @@
                       (loc
                        ((loc_start
                          ((pos_fname generated_locations.ml) (pos_lnum 1)
-                          (pos_bol 0) (pos_cnum 20)))
+                          (pos_bol 0) (pos_cnum 18)))
                         (loc_end
                          ((pos_fname generated_locations.ml) (pos_lnum 1)
-                          (pos_bol 0) (pos_cnum 23)))
+                          (pos_bol 0) (pos_cnum 65)))
```
jchavarri, line 758: These are pretty big changes, do you know what happened?

ppx/src/reactjs_jsx_ppx.ml
```diff
@@ -1,15 +1,88 @@
...
+
+    let array ~loc children =
+      Builder.pexp_apply ~loc
+        (Builder.pexp_ident ~loc
+           { txt = Longident.Ldot (Lident "React", "array"); loc })
+        [ (nolabel, children) ]
+
+    let componentLike ~loc props return =
+      Ptyp_constr
+        ( { loc; txt = Ldot (Lident "React", "componentLike") },
```
jchavarri, line 59: This and `ReactDOM.domProps` are types, not bindings. But I guess it is ok :)

jchavarri (approved):

ppx/src/reactjs_jsx_ppx.ml
davesnx: I guess it's fine, right?

Maybe Bindings.Types.* is an overkill?

Issue #770, opened 2023-09-06
Wrap the React Library


anmonteiro:
reason-react exposes too many toplevel modules. These could be nested under `React`, e.g.:

- `React.DOM`
- `React.ErrorBoundary`
- `React.DOMServer`
etc

anmonteiro:
I don't think that's a good idea. Why is it so important that our library structure matches the JS package names / sub-paths?

I actually think that the current wrapped design is quite elegant and there's a nice sub-module for everything that you can access.

davesnx:
> I don't think that's a good idea. Why is it so important that our library structure matches the JS package names / sub-paths?

First for discoverability and second to have the minimum amount of subjectivity on the names of the structure.

anmonteiro:
I don't think it's very discoverable having to add `(libraries react react.dom react.dom-server)` in order to have those modules appear in your editor. It's a much better API to have `(libraries reason-react)` and `React.DOM` etc. don't you think?

davesnx:
What I'm proposing is to have `libraries reason-react` which exposes react, reactdom, etc. and not fall into the trap what we had before so, `React.Event`, `ReactDOM.Style`, `ReactDOM.Server` etc still remains wrapped.

anmonteiro:
how do you expose those multiple toplevel modules from:

1. a single library `reason-react`
2. without exposing more?

davesnx:
`wrapped false`, one module per toplevel, the rest either inlined or private libraries

anmonteiro:
That actually sounds pretty nice if you can make it work. I'm in favor of it, if can manage to hide the internal modules.

PR #773, opened 2023-09-06
feat: reduce too many exposes module in toplevel

r17x:
Based on this #770


davesnx:
Instead of going to that direction, probably `(wrapped false)` is a smarter move?

r17x:
@davesnx  I'm not familiar with term of `(wrapped false)` could you explain about it?

davesnx:
`wrapped false` is the default behaviour from ReScript where each module is exposed so given a library like reason-react you will expose all modules (React, ReactDOM, ReasonReactRouter, etc) 

With `wrapped true` you will expose this library under a name (or other more exotic forms) so lets say I name this library `React`, I could expose all modules under React now: (React.React, React.ReactDOM, React.ReasonReactRouter, etc)

anmonteiro:
> Instead of going to that direction, probably (wrapped false) is a smarter move?

We'll need both. This PR alone still exposes all the other modules because there's `(wrapped false)` in the dune file.

src/dune
```diff
@@ -5,5 +5,5 @@
  (public_name reason-react)
  (preprocess
   (pps melange.ppx reason-react-ppx))
- (wrapped false)
+ (wrapped true)
```
r17x, line 8: I have been changes to be `true` @davesnx . I think is incomplete yet? could you give me suggestions for this changes?

anmonteiro: One way to move forward with this would be to fix the build

anmonteiro:
@davesnx do you want to include this for 0.12, or should it be included after that?

davesnx:
Let's do it for 0.13

anmonteiro:
This is actually going to be a lot harder than initially thought, since many modules depend on the `React` module.

anmonteiro:
solution in #783

r17x:
I learn a lot from #783 and I closed this PRs

Good job!

Thanks @anmonteiro

PR #774, opened 2023-09-07
opam: correctly mark some deps as build

rizo:
Hey! Started using reason-react and noticed that some deps aren't properly marked as `build` which causes issues with nix.

cc @leojrfs


davesnx:
Thanks for the change, I believe it should change on the dune-project also!

anmonteiro:
These are not build dependencies. They can cause issues when downgrading. 

This is a good summary of the problem: https://github.com/ocaml/opam/issues/3850#issuecomment-882632871

If you look at opam repository, `dune` is explicitly never marked as a `{build}` dependency.

PR #781, opened 2023-09-14
Make sure opam files have the same shape as opam repository after PR

davesnx:
Updating opam files after https://github.com/ocaml/opam-repository/pull/24422/files

Thanks @haochenx

### DIFF
--- a/reason-react-ppx.opam
+++ b/reason-react-ppx.opam
@@ -22,6 +22,7 @@ depends: [
   "ocamlformat" {= "0.24.0" & with-dev-setup}
   "odoc" {with-doc}
 ]
+dev-repo: "git+https://github.com/reasonml/reason-react.git"
 build: [
   ["dune" "subst"] {dev}
   [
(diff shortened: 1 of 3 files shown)


anmonteiro:
TBH I don't think this is necessary. We can just strip `@runtest` when releasing to opam. That's what I do in some of my packages too.

Doesn't justify having `.opam.template` files in the repo IMO.

davesnx:
What's the problem with the opam template? I added to avoid the manual work on each release, which is rather annoying

haochenx:
(since I'm pinged, allow me to add my two cents)

It's probably better (and safer?) to let Dune automatically generate the build command. I think @anmonteiro means stripping `@runtest` is probably a trivial diff to lose the benefits of having the Dune generated build commands.

Also, if the approach in https://github.com/ocaml/opam-repository/pull/24422/files#r1325843840 is adopted, then the stripping should not be necessary anymore.

(btw sadly the `@runtest` is hardcoded in Dune for now. ref: https://github.com/ocaml/dune/blob/main/src/dune_rules/opam_create.ml)

davesnx:
That makes a lot of sense. Will try to use `{post}` on the next release and avoid the templates

haochenx:
> Will try to use {post} on the next release

Just for your information: you can indeed open an PR to opam-repository to modify the current version to test whether `{post}` works now. opam-repository can take amendments to existing packages (and we see chose "maintenance" patches quite often)

PR #786, opened 2023-09-18
fix: remove `[@mel.uncurry]` from `React.useCallback*` functions

anmonteiro:
`React.useCallback` can take a function of multiple arguments. In the following example:

```reason
let foo = React.useCallback((_a, _b) => Js.log("HELLO"));
```

The logging never occurs with `[@mel.uncurry]` because Melange generates the following code:

```js
function (_a) { return function (_b) { ... }}
```

We preserve the uncurried callbacks in `React.Uncurried.useCallback*`


jchavarri:
> In the following example:

```reason
let foo = React.useCallback((_a, _b) => Js.log("HELLO"));
```

Maybe this could be added to the tests.

jchavarri:
I tried to understand better why this change works, and why it doesn't break any other existing cases. Leaving here a note for the record.

The reason is that `uncurry` is only necessary if the callback function is called from the JavaScript side. For `useCallback`, this never happens: React doesn't check or call the function itself, nor the # of arguments it has. In other words, the function will only be called from the OCaml side.

> We preserve the uncurried callbacks in `React.Uncurried.useCallback*`

The above makes me wonder about the `Uncurried` version. If we know the function will always be called from OCaml side, is there a point to keep the `mel.uncurry` attribute in that one?

anmonteiro:
> For useCallback, this never happens: React doesn't check or call the function itself, nor the # of arguments it has. In other words, the function will only be called from the OCaml side.

This isn't true. I caught the bug when passing it to JavaScript, e.g.:

```reason
module X = {
  [@mel.module "foobar"] [@react.component]
  external make: (~takesFunction: (a, b) => unit) => React.element = "default";
};

module FromOCaml = {
  [@react.component]
  let make = () => {
    let myFn = React.useCallback0((a,b)=>a+b);
    <X takesFunction=myFn />;
  };
};

**Comment by jchavarri** [comment]
I think in the case you mention, the `takesFunction` prop should be using `@u` attribute to uncurry. Otherwise it can be trivially broken. E.g. what happens if you do this?

```
let myFn = React.useCallback0((a) => {let f = (b) =>a+b; f});
<X takesFunction=myFn />;
```

I think it breaks even without `useCallback`, so the problem does not seem related to `React.useCallback` at all?

**Comment by anmonteiro** [comment]
That's true. Are you suggesting we remove `[@mel.uncurry]` from the `Uncurried` module too?

**Comment by anmonteiro** [comment]
It's also not possible to use uncurried functions in `useCallback`. [Playground example](https://melange.re/unstable/playground/?language=Reason&code=dHlwZSBjYWxsYmFjaygnYSwgJ2IpID0gJ2EgPT4gJ2I7CgpbQG1lbC5tb2R1bGUgInJlYWN0Il0KZXh0ZXJuYWwgdXNlQ2FsbGJhY2swOgogICgoJ2lucHV0ID0%2BICdvdXRwdXQpLCBbQG1lbC5hcyB7anNvbnxbXXxqc29ufV0gXykgPT4KICBjYWxsYmFjaygnaW5wdXQsICdvdXRwdXQpID0KICAidXNlQ2FsbGJhY2siOwoKbGV0IHggPSAoKSA9PiB7CiAgdXNlQ2FsbGJhY2swKCguIGEsIGIpID0%2BIGEgKyBiKTsKfTsK&live=off):

image

**Comment by anmonteiro** [comment]
So perhaps the `Uncurried` module should specify those explicitly? Something like this (better naming suggestions appreciated):

```reason
module Uncurried = {
  [@mel.module "react"]
  external useCallback1_0:
    ((. 'input) => 'output, [@mel.as {json|[]|json}] _) =>
    callback('input, 'output) =
    "useCallback";

  [@mel.module "react"]
  external useCallback2_0:
    ((. 'input1, 'input2) => 'output, [@mel.as {json|[]|json}] _) =>
    ((. 'input1, 'input2) => 'output) =
    "useCallback";
};```

**Comment by jchavarri** [comment]
> Are you suggesting we remove `[@mel.uncurry]` from the `Uncurried` module too?

Not really. With the [message above](https://github.com/reasonml/reason-react/pull/786#issuecomment-1725117792) I just wanted to express that the code in the `react` JS library (which we bind against) will never call  the function passed to `useCallback`. So using `uncurry` or not in this case does not make any difference as you noted originally :)

I am not sure about the support for uncurried functions... I need to think about it  more.

**Comment by jchavarri** [comment]
I think we could simplify things quite a bit if we get rid of the function type on the `useCallback` binding. So instead of `'input => 'output` we just use `'a`.

I know this might sound unsafe, but ReactJS itself doesn't care about whatever is on the value passed to `useCallback`. And from OCaml side, things will be type safe because whatever gets passed to `useCallback` comes through the other side. You can potentially call `useCallback(5)` but you could never "call" it.

[Playground](https://melange.re/unstable/playground/?language=Reason&code=W0BtZWwubW9kdWxlICJyZWFjdCJdCmV4dGVybmFsIHVzZUNhbGxiYWNrMDogKCdhLCBbQG1lbC5hcyB7anNvbnxbXXxqc29ufV0gXykgPT4gJ2EgPSAidXNlQ2FsbGJhY2siOwpsZXQgeCA9IHVzZUNhbGxiYWNrMCgoLiBhLCBiKSA9PiBhICsgYik7CgpsZXQgeSA9IHgoMik7CmxldCB5ID0geCguIDIsIDMpOwo%3D&live=off).

**Comment by anmonteiro** [comment]
Seems more correct than the current situation anyway.

**Comment by anmonteiro** [comment]
@jchavarri used your suggestion. Added a generic `'fn` argument to the `useCallback` functions, and removed them from the `Uncurried` module, since they'd be the same.

File: src/React.re [hunk 1]
```diff
@@ -346,51 +346,24 @@ external useMemo7:
...
-external useCallback6:
-  ([@mel.uncurry] ('input => 'output), ('a, 'b, 'c, 'd, 'e, 'f)) =>
-  callback('input, 'output) =
-  "useCallback";
+external useCallback6: ('fn, ('a, 'b, 'c, 'd, 'e, 'f)) => 'fn = "useCallback";
 [@mel.module "react"]
-external useCallback7:
-  ([@mel.uncurry] ('input => 'output), ('a, 'b, 'c, 'd, 'e, 'f, 'g)) =>
-  callback('input, 'output) =
+external useCallback7: ('fn => 'output, ('a, 'b, 'c, 'd, 'e, 'f, 'g)) => 'fn =
```
davesnx, line 366: `=> 'output` should be removed

anmonteiro:
> It's minor but there's a runtime check of being sure that useCallback gets a function. That's probably why the callback type was added in the first place.

Do you mean in React.js proper? That's even better, then. React can make sure we're passing a function (though it'll be caught at development time).

CHANGES.md
```diff
@@ -2,6 +2,8 @@
 
 * Remove legacy `ReactDOMRe` and `ReasonReact` modules (@anmonteiro in
   [#782](https://github.com/reasonml/reason-react/pull/782))
+* Fix `React.useCallback*` for callbacks with multiple arguments (@anmonteiro
```
jchavarri, line 5: Should this be now "for uncurried callbacks"?

jchavarri (approved):
Looks good afaict, hopefully the weaker types don't pose an issue. I was trying to trigger the warning that @davesnx mentioned, but I was not able to.

CHANGES.md
anmonteiro: I think it's both, so best to leave it unspecified.

jchavarri:
Seems React just swallows whatever is passed to it: https://stackblitz.com/edit/react-playground-practice-13ux64?file=index.js

test/blackbox-tests/useCallback.t
```diff
@@ -0,0 +1,53 @@
...
+  $ cat >dune <<EOF
+  > (melange.emit
+  >  (target js-out)
+  >  (emit_stdlib false)
+  >  (preprocess (pps melange.ppx))
+  >  (libraries reason-react))
+  > EOF
+
+  $ cat >x.ml <<EOF
+  > let cb = React.useCallback0(fun a b -> a + b)
```
jchavarri, line 16: Sorry for late feedback. Just realized: do we want to document the case I mentioned before? the one that produces a curried function in JS: `fun a -> let f b = a + b in f`.

anmonteiro: I don't think so. That's quite obvious from the way it's written.

davesnx:
@jchavarri I got a TS error on JSX and didn't realised useCallback eats everything.

PR #791, opened 2023-09-26
fix: option type in external generation

anmonteiro:


ppx/test/signature.t/run.t
```diff
@@ -49,7 +49,7 @@
   module MyPropIsOptionOptionBool = {
     [@mel.obj]
     external makeProps:
-      (~myProp: option(bool)=?, ~key: string=?, unit) =>
+      (~myProp: bool=?, ~key: string=?, unit) =>
       {. "myProp": option(option(bool))};
```
davesnx, line 53: This doesn't look good, myProp shouldn't be `option(option ...`

ppx/reason_react_ppx.ml
```diff
@@ -321,18 +321,15 @@ let rec recursivelyMakeNamedArgsForExternal list args =
            | _label, Some type_, Some _ -> type_
            (* ~foo: option(int)=? *)
            | ( _label,
-               Some
```
davesnx, line 324: This case should remain here and add a discrimination when it's not `*predef*`, otherwise the wrapping from the user disapears

davesnx (changes requested):

ppx/test/signature.t/run.t
anmonteiro: I don't get it? this looks correct in the signature part.

ppx/reason_react_ppx.ml
anmonteiro: In regular OCaml, and I'll include a [playground link](https://melange.re/v2.0.0/playground/?language=Reason&code=bGV0IG15Rm4gPSAofmE6IG9wdGlvbihpbnQpPT8sICgpKSA9PiBPcHRpb24udmFsdWUoYSwgfmRlZmF1bHQ9MCk7CgpleHRlcm5hbCBteUZuRXh0ZXJuYWw6ICh%2BYTogaW50PT8sIHVuaXQpID0%2BIGludCA9ICJkb2VzbkV4aXN0IjsKCmxldCAoKSA9IHsKICBsZXQgXzogaW50ID0gbXlGbih%2BYT0zLCAoKSk7CiAgbGV0IF86IGludCA9IG15Rm5FeHRlcm5hbCh%2BYT0zLCAoKSk7CgogIGxldCBfOiBpbnQgPSBteUZuKCk7CiAgbGV0IF86IGludCA9IG15Rm5FeHRlcm5hbCgpOwoKICBsZXQgXzogaW50ID0gbXlGbih%2BYT0%2FU29tZSgzKSwgKCkpOwogIGxldCBfOiBpbnQgPSBteUZuRXh0ZXJuYWwofmE9P1NvbWUoMyksICgpKTsKICAoKTsKfTsK&live=off), 

there's the following distinction:

```reason
let myFn = (~a: option(int)=?, ()) => Option.value(a, ~default=0);

external myFnExternal: (~a: int=?, unit) => int = "doesnExist";
```

these have the same signature. my change makes our PPX consistent with the default OCaml behavior. I believe this is fixing a bug. What use cases am I missing?

anmonteiro:
@davesnx the real difference for inferring the type for the `makeProps` external needs to consider whether the React component is being defined by a `let` or an `external`. In the former, optionally labeled arguments need to be annotated with `~foo: option('a)=?` whereas in externals / signatures they're rather `~foo: 'a=?`. The PPX now considers that. I've added more examples.

ppx/test/signature-optional.t/input.re
```diff
@@ -0,0 +1,42 @@
...
+  external make: (~myProp: option(bool)=?) => React.element = "B";
+};
+
+module MyPropIsOptionOptionBoolLetWithValSig: {
+  [@react.component]
+  let make: (~myProp: option(bool)=?) => React.element;
+} = {
+  [@react.component]
+  let make = (~myProp: option(option(bool))=?) => React.null;
+};
```
davesnx, line 42: That's extremely helpful <3

davesnx (approved):

ppx/reason_react_ppx.ml
```diff
@@ -1093,8 +1097,8 @@ let jsxMapper =
             let namedTypeList = List.fold_left argToType [] namedArgList in
             let loc = gloc in
             let externalDecl =
-              makeExternalDecl fnName loc namedArgListWithKeyAndRef
-                namedTypeList
+              makeExternalDecl ~component_is_external:false fnName loc
```
jchavarri, line 1100: We might want to add `component_is_external` to `makeExternalDecl` declaration params to avoid confusing errors with currying down the line.

jchavarri (approved):

PR #792, opened 2023-09-30
fix method attr name

joprice:
Uses of the `method` attribute were producing `method_` in the js output. This adds a mel.as attribute to avoid the escaping of the keyword when converting from reason to ml.

### DIFF
--- a/src/React.rei
+++ b/src/React.rei
@@ -1820,7 +1820,7 @@ module DOM: {
     media: option(string), /* a valid media query */
     [@mel.optional]
     mediaGroup: option(string),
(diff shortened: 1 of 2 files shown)


joprice:
@anmonteiro filled me in that I'm hitting this issue because of the version of reason I'm using from the [nix-overlays](nix-ocaml/nix-overlays: OCaml-focused, custom nix-overlays), which has this change reverted https://github.com/reasonml/reason/pull/2530. I wonder if this change is still valid, if the `mel.as` will have the same effect in either context, whether the reason code maps it to `method` or `method_`.

PR #794, opened 2023-10-11
Re-organise the modules

davesnx:
As discussed here: https://github.com/reasonml/reason-react/issues/770

This file explains the new organisation of modules: https://github.com/reasonml/reason-react/blob/Unwrap-and-minimize-toplevels/src/README.md

- `wrapped false` and inlining most of the modules makes the interface a little cleaner for the users. On the other side, we have huge modules (which isn't so bad, but that it just my opinion)

- Moved all interface from `React.rei` into it's own module

- I didn't took any decision on `ErrorBoundary`, neither `ReasonReactRouter`. I got the feeling that we should rename those as `SimpleErrorBoundary` and `SimpleRouter` or similar. I would be fine to rename `ReasonReactRouter` to `Router`


src/README.md
```diff
@@ -1,17 +1,22 @@
-Welcome to the source of ReasonReact!
+### Welcome to the source of ReasonReact
+
+We want to expose as minimum modules to the toplevel as possible, and at the same time want to map closely to the npm packages. So each module maps to a npm package, and each sub-module maps to a sub-module. For example: `react-dom` -> `ReactDOM` and `react-dom/server` -> `ReactDOM.Server`.
 
 Files overview:
 
 ## Bindings
```
anmonteiro, line 7: thanks for writing this up. it's very clear. 

you may want to check the rendering before merging, as the nested bullets aren't rendered as you'd expect.

src/README.md
```diff
@@ -1,17 +1,22 @@
...
+  |- `ReactDOM.Style`: bindings to create `style` objects
+  |- `ReactDOM.TestUtils`: helpers for testing your components
 
-## Extra
+## Extra (not part of React)
 
-- `ReactTestUtils`: helpers for testing your components
-- `ReasonReactErrorBoundary`: component to catch errors within your component tree
 - `ReasonReactRouter`: a simple, yet fully featured router with minimal memory allocations
+- `ErrorBoundary`: component to catch errors within your component tree
```
anmonteiro, line 20: should this one become `React.ErrorBoundary`, or do you want it to stay out at the toplevel?

src/React.re
```diff
@@ -139,10 +456,428 @@ module Suspense = {
...
+[@mel.module "react"]
+external useMemo6:
+  ([@mel.uncurry] (unit => 'any), ('a, 'b, 'c, 'd, 'e, 'f)) => 'any =
+  "useMemo";
+[@mel.module "react"]
+external useMemo7:
+  ([@mel.uncurry] (unit => 'any), ('a, 'b, 'c, 'd, 'e, 'f, 'g)) => 'any =
+  "useMemo";
+
+/* This is used as return values  */
```
anmonteiro, line 693: This was for `type callback('input, 'output)`. which is no longer here

```suggestion
/* This is used as return values  */
```

src/React.rei
```diff
@@ -195,7 +195,7 @@ external useSyncExternalStore:
 [@mel.module "react"]
 external useSyncExternalStoreWithServer:
   (
-    ~subscribe: (([@mel.uncurry] (unit => unit)) => ([@mel.uncurry] (unit => unit))),
+    ~subscribe: ([@mel.uncurry] (unit => unit)) => [@mel.uncurry] (unit => unit),
```
anmonteiro, line 198: refmt did this?

src/dune
```diff
@@ -1,6 +1,10 @@
 (library
  (name react)
  (public_name reason-react)
+ (wrapped false)
+ ; Explicitly adding modules isn't necessary, but it's a good idea
```
anmonteiro, line 5: great idea 👍

anmonteiro (commented):
looking very nice

src/React.rei
davesnx: Yes

src/React.re
davesnx: Removed this and added into the React.Uncurried callback

src/React.rei
davesnx: Tested and looks fine: https://melange.re/unstable/playground/?language=Reason&code=W0BtZWwubW9kdWxlICJyZWFjdCJdCmV4dGVybmFsIHVzZVN5bmNFeHRlcm5hbFN0b3JlOgogICgKICAgIH5zdWJzY3JpYmU6IChbQG1lbC51bmN1cnJ5XSAodW5pdCA9PiB1bml0KSkgPT4gW0BtZWwudW5jdXJyeV0gKHVuaXQgPT4gdW5pdCksCiAgICB%2BZ2V0U25hcHNob3Q6IHVuaXQgPT4gJ3NuYXBzaG90LAogICAgfmdldFNlcnZlclNuYXBzaG90OiBbQG1lbC51bmN1cnJ5XSAodW5pdCA9PiAnc25hcHNob3QpCiAgKSA9PgogICdzbmFwc2hvdCA9CiAgInVzZVN5bmNFeHRlcm5hbFN0b3JlIjsKCmxldCBnZXRTdGF0ZSA9ICgpID0%2BIDIyOwpsZXQgc3Vic2NyaWJlV2l0aE1vY2sgPSBsaXN0ZW5lciA9PiB7CiAgbGV0IF8gPSBsaXN0ZW5lcigpOwogIEpzLmxvZygiY29zaXRhcyIpOwogICgpID0%2BICgpOwp9OwoKbGV0IHNuYXBzaG90ID0KICB1c2VTeW5jRXh0ZXJuYWxTdG9yZSh%2Bc3Vic2NyaWJlPXN1YnNjcmliZVdpdGhNb2NrLCB%2BZ2V0U25hcHNob3Q9Z2V0U3RhdGUpOwo%3D&live=off

src/README.md
anmonteiro: OK then I'm fine taking the time to think about this, and keeping it `ReasonReactErrorBoundary` for now?

anmonteiro (approved):
decide what to name `errorBoundary.re` -- I propose `ReasonReactErrorBoundary` to give it a long, unique name.

thanks for working on this

src/README.md
davesnx: Sounds perfect

davesnx:
We would need to make a proper release notes with the module changes, but should be small.

joprice:
@davesnx I'm seeing react-dom/test-utils and react-dom/server getting pulled in by vite during development with this change. Without some sort of change in melange to only import referenced submodules, maybe new modules like `ReactDOMServer` and `ReactDOMTestUtils` are necessary to avoid this?

davesnx:
Splitting those into modules might make sense to avoid this issue. Would this problem fade away in production builds, where those modules would be gone with tree-shaking?

I would like to now how other bundlers handle dev/prod optimisations like this one, to decide if it's a good solution

joprice:
I didn't get that far, since vite didn't like the `require` calls in the server-side commonjs modules. Not sure these particular ones will be shaken away, since I think purity of the module affect tree-shaking as well and would have to check that. I think either way, it's unexpected to pull in test utils or, in general, something you didn't intend to import, even in dev.

PR #776, opened 2023-09-07
Make optional props optional properly

davesnx:
Fixes https://github.com/reasonml/reason-react/issues/618

PR #799, opened 2023-10-19
fix: remove redundant check related to option(_) labelled props

anmonteiro:
I don't think this check is necessary anymore, since #776 and #791. 

After those PRs we infer the types correctly for optionally labelled arguments.

### DIFF
--- a/ppx/reason_react_ppx.ml
+++ b/ppx/reason_react_ppx.ml
@@ -639,29 +639,6 @@ let jsxMapper =
           default,
           pattern,
           expression ) ->
-        let () =
-          match (arg, pattern, default) with
-          | Optional _, { ppat_desc = Ppat_constraint (_, { ptyp_desc }) }, None -> (
-              match ptyp_desc with
-              | Ptyp_constr ({ txt = Lident "option" }, [ _ ]) -> ()
-              | _ ->
-                  let currentType =
-                    match ptyp_desc with
-                    | Ptyp_constr ({ txt }, []) ->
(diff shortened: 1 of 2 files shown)


ppx/reason_react_ppx.ml
```diff
@@ -641,29 +641,6 @@ let jsxMapper =
...
-                    | Ptyp_constr ({ txt }, []) ->
-                        String.concat "." (Longident.flatten_exn txt)
-                    | Ptyp_constr ({ txt }, _innerTypeArgs) ->
-                        String.concat "." (Longident.flatten_exn txt) ^ "(...)"
-                    | _ -> "..."
-                  in
-                  Ocaml_common.Location.prerr_warning pattern.ppat_loc
-                    (Preprocessor
-                       (Printf.sprintf
-                          "react-jsx-ppx: optional argument annotations must \
```
jchavarri, line 661: Should we first add some tests that exercise this code path, then apply this change to see how they are affected? it's impossible for me to know what this change is doing by just looking at the code. My understanding is that it was making sure that optional labelled args in props are properly annotated, right?

anmonteiro: Adding tests is a great idea. Will do that to help us visualize the coverage better 👍

anmonteiro:
@jchavarri added tests and re-ordered the commits to show how the check was redundant.

jchavarri (approved):
Thanks! it is now super clear what the change does.

PR #802, opened 2023-10-20
testrenderer: add toObject

jchavarri:
Follow up to https://github.com/reasonml/reason-react/pull/797.

### DIFF
--- a/src/ReactTestRenderer.re
+++ b/src/ReactTestRenderer.re
@@ -4,7 +4,7 @@ type t;
 external create: React.element => t = "create";
 
 [@mel.send] external toJSON: t => Js.Json.t = "toJSON";
-[@mel.send] external toObject: t => Js.t({.}) = "%identity";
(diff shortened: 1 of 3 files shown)


src/ReactTestRenderer.re
```diff
@@ -4,6 +4,7 @@ type t;
 external create: React.element => t = "create";
 
 [@mel.send] external toJSON: t => Js.Json.t = "toJSON";
+[@mel.send] external toObject: t => Js.t({.}) = "%identity";
```
anmonteiro, line 7: should probably be `Js.t({..})` to allow getting any key out with `##`?

jchavarri: totally, fixed in 4940103.

anmonteiro: You forgot to change the interface file, which is where it really matters in this case .

PR #805, opened 2023-10-23
fix refs to reactdom

jchavarri:


src/ReactDOM.rei
```diff
@@ -510,13 +510,13 @@ module Client: {
 };
 
 [@deprecated
-  "This function will be removed in the next release. Use React.DOM.Client.render instead."
+  "This function will be removed in the next release. Use ReactDOM.Client.render instead."
```
jchavarri, line 513: Not sure if we should mention `ReactDOM.Client.createRoot` here as well, I always have to look around to see where it has been placed.

anmonteiro (approved):

src/ReactDOM.rei
davesnx: Totally

jchavarri: https://github.com/reasonml/reason-react/pull/805/commits/c9f1b7b1b6f4e1f2e143589ffbf7c8d1e08db71b

PR #807, opened 2023-10-24
Proposal: native compat library for syncing with server-reason-react

jchavarri:
As [server-reason-react](https://github.com/ml-in-barcelona/server-reason-react/) is mirroring the API of `reason-react`, it becomes challenging to keep up over there with the changes done here.

We also don't want to remove the `external` declarations used on `react.rei` for performance reasons.

So the solution is to create a second `react.rei` that is used purely for syncing purposes with `server-reason-react`, until we have virtual libraries or better support for this. See companion PR https://github.com/ml-in-barcelona/server-reason-react/pull/82.

### DIFF
--- a/src/native_compat/React.rei
+++ b/src/native_compat/React.rei
@@ -137,11 +137,15 @@ let useReducerWithMapState:
   ('state, 'action => unit);
 
 let useSyncExternalStore:
-  (~subscribe: unit => unit, ~getSnapshot: unit => 'snapshot) => 'snapshot;
+  (
+    ~subscribe: (unit => unit, unit) => unit,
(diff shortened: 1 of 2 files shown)


jchavarri:
Talked to @davesnx offline and we decided to explore a different path, where we use `server-reason-react.melange.ppx` which is already removing the differences between melange and native, to process `react.rei` and convert externals into regular `let` definitions, e.g.

```
[@mel.module "react"]
external useMemo: ([@mel.uncurry] (unit => 'any)) => 'any = "useMemo";
```

into
```
let useMemo: (unit => 'any) => 'any;
```

That way, we might be able to start working towards a reason-react virtual library that includes the mli "as is" currently.

PR #811, opened 2023-11-02
Update React.re

davesnx:


anmonteiro:
What's the goal of this?

davesnx:
My bad, I was the type was repeated, but now I see there's a uncurried version and a curried version

PR #815, opened 2023-11-07
Move pipeablestream to new lib

### LINKED ISSUE #814: Bundling errors in ES6 due to `react-dom` having different API in Node and browser
When trying to use the latest `reason-react` with react 18 using es6 modules in the browser, the bundling step fails with:

```
WARNING in ../../../_build/default/frontend/all/node_modules/reason-react/ReactDOM.bs.js 59:9-38
export 'renderToPipeableStream' (imported as 'Server') was not found in 'react-dom/server' (possible exports: renderToNodeStream, renderToReadableStream, renderToStaticMarkup, renderToStaticNodeStream, renderToString, version)
```

I suspect a similar issue happens if one tries to use it from Node, but in that case `renderToReadableStream` will fail as it's not available in Node.

The reasons are:
- `react-dom` doesn't have uniform APIs across browser and node implementations.
- es6 imports are analyzed statically, so even if the `renderToPipeableStream` is never called, the resolver tries to solve the import at build time (that's why commonjs works)

A solution could be to move the offending APIs `renderToPipeableStream` and `renderToReadableStream` to new libraries `react-dom.node-only` or `react-dom.browser-only`.

Ref: https://github.com/facebook/react/issues/24226.

jchavarri:
Fixes #814.

Adds a new library `reason-react.node-only` that only includes the binding to `renderToPipeableStream` and companion types and utilities.


davesnx (approved):
LGTM probably `reason-react.node` looks good too!

anmonteiro:
Why don't we make `ReactDOMServer` itself the entire separate library?

anmonteiro:
Check https://github.com/reasonml/reason-react/pull/819 with a proposal.

I'm not a big fan of `node-only` since that also works in deno, bun, other runtimes.

Instead, my proposal separates the entire `ReactDOMServer` module to a new `reason-react.dom-server` library. 

Let me know what you think.

PR #819, opened 2023-11-08
chore: rename ReactDOMServerNode{Only,}

anmonteiro:
follow-up to https://github.com/reasonml/reason-react/pull/815


jchavarri (commented):
What problem is this change solving?

I think there might be more limitations related to #814 down the line on because react-dom has different implementations for Node, Bun, Browser, Edge and they might diverge more over time. To solve that, I think we can continue adding specific modules for the missing engines.

Adding a new library means that ReasonReact users that don't need `renderToPipeableStream` can not keep using ReasonReact without having to add two libraries as deps.

Maybe I'm missing st?

davesnx:
I liked node becaues renderToPipeableStream just works in node, eventually splitting react and react-dom seems innevitable in reason-react

I got the feeling we spend too much time on renaming things on all clients of reason-react that any change is really painful for just us.

jchavarri:
Discussed offline.

Maybe we can look at it from the docs perspective https://react.dev/reference/react-dom/server.

If we follow that arrangement, it would mean we would expose 3 libs:

- node
- web
- non-streaming

We can expand with bun or whatever in the future if it also has different API (I suspect it will follow Node).

anmonteiro:
I wasn't aware that `renderToPipeableStream` only worked in Node, or about the state of `react-dom/server` and its different implementations.

In that case, I like Javi's idea.

anmonteiro:
This PR is now just changing the name of `ReactDOMServerNodeOnly` to `ReactDOMServerNode`. `NodeOnly` felt like a weird name, but I don't have a strong opinion.

jchavarri (approved):

PR #821, opened 2023-11-15
Melange v3 long-lived branch

anmonteiro:


ppx/test/issue-429.t/run.t
```diff
@@ -44,7 +44,7 @@ Let's test hovering over parts of the component
         "line": 36,
         "col": 1
       },
-      "type": "(~greeting: string) => React.element",
+      "type": "(~greeting: string) => element",
```
jchavarri, line 47: These changes are strange, I assume is the new merlin version used? I can't see anything in reason-react itself that changed, or in melange.

ppx/test/keys.t/run.t
```diff
@@ -25,14 +25,14 @@ _^
...
-      "col": 0
+      "line": 10,
+      "col": 2
     },
     "end": {
       "line": 10,
       "col": 85
     },
-    "type": "{.. \"author\": Author.t} => React.element",
+    "type": "string",
```
jchavarri, line 35: This is also strange, but I guess now it's more accurate (I think this test is hovering over `key` prop, from what I see in line 21).

jchavarri (approved):
Looks good, thanks! I'm a bit confused about the changes in locations, and why I updated the merlin version at some point, I assume there was some incompatibility with Melange v3 or other package 🤔

ppx/test/issue-429.t/run.t
anmonteiro: I don't understand why these changed, but they're both correct and consistent across opam / nix calls, so I didn't worry too much.

anmonteiro:
I went ahead and restored the branch here. If your builds somehow failed because this melange-v3 branch was missing, please retry the build!

PR #823, opened 2023-11-21
ppx: support "custom children" in uppercase components without having to wrap in array literal

### LINKED ISSUE #822: Support "custom children" in uppercase components
Consider the following component:

```reason
module Test = {
  type t = {foo: int};
  [@react.component]
  let make = (~children) => {
    let () = Array.iter(c => Js.log(c.foo), children);
    <div />;
  };
};
```

An a usage:

```reason
let el = <Test> {foo: 2} {foo:3} </Test>
```

This fails with

```
11 |   ((Test.createElement ~children:[{ Test.foo = 2 }; { Test.foo = 3 }] ())
                                       ^^^^^^^^^^^^^^^^
Error: This expression has type Test.t but an expression was expected of type React.element
```

The reason is that the PPX compiles the above into:

```reason
let el =
  React.jsxs(
    Test.make,
    Test.makeProps(
      ~children=React.array([|{Test.foo: 2}, {Test.foo: 3}|]),
      (),
    ),
  );
```

And `React.array` expects (correctly) an array of elements as input.

I wonder if the PPX could skip the call to `React.array` for elements that are produced using custom components (i.e. the upper case ones)? In those cases, the component author has full control, so if they pass an array, we can allow the type inside the array to be anything, really.

jchavarri:
Fixes #822.

The breaking change is for uppercase components that were introspecting with `React.Children`. After the change, these components would need to run `children` through `React.array`.

It allows to write:

```reason
<Test> {Test.name: "foo"} {name: "bar"} </Test>
```

besides the already possible:

```reason
<Test> [|{Test.name: "foo"}, {name: "bar"}|] </Test>
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+* BREAKING, ppx: Allow passing an array of custom children to a component
+  without having to wrap in array literal ([@jchavarri in #748](https://github.com/reasonml/reason-react/pull/823))
+
 # 0.15.0
 
 * Add `isValidElement` (@r17x in
--- a/ppx/reason_react_ppx.ml
+++ b/ppx/reason_react_ppx.ml
@@ -458,7 +458,14 @@ let makePropsType ~loc namedTypeList =
       };
     ]
 
-let jsxExprAndChildren ~ident ~loc ~ctxt mapper ~keyProps children =
+type component_type = Uppercase | Lowercase
+
+let jsxExprAndChildren ~component_type ~loc ~ctxt mapper ~keyProps children =
+  let ident =
+    match component_type with
+    | Uppercase -> Lident "React"
+    | Lowercase -> Lident "ReactDOM"
+  in
   let childrenExpr =
     Option.map (transformChildrenIfListUpper ~loc ~mapper ~ctxt) children
   in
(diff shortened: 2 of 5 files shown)


anmonteiro (approved):
I think this make sense. 

I wonder if we should mention this difference between uppercase / lowercase components in our documentation.

jchavarri:
@davesnx what are your thoughts on this?

> I wonder if we should mention this difference between uppercase / lowercase components in our documentation.

@anmonteiro We do mention something already:
- https://reasonml.github.io/docs/en/jsx#capitalized-tag
- https://reasonml.github.io/docs/en/jsx#uncapitalized-tag

Do you mean to expand on that?

anmonteiro:
> @anmonteiro We do mention something already:
> 
> * https://reasonml.github.io/docs/en/jsx#capitalized-tag
> * https://reasonml.github.io/docs/en/jsx#uncapitalized-tag
> 
> Do you mean to expand on that?

Right, I wonder if we should mention that these are no longer wrapped in array, or if it's OK just to document it in the changelog.

davesnx (approved):
Looks like a great idea, uppercase components supporting any children looks nice

anmonteiro:
merging this as discussed in https://ahrefs.slack.com/archives/C01RF16D2LA/p1732103117043289

let's revert if anyone disagrees.

PR #824, opened 2023-11-22
test: add to runtest + fix bindings

jchavarri:
- ~Runs `jest` as part of `runtest` alias~ Edit: see [below](https://github.com/reasonml/reason-react/pull/824#issuecomment-1823087705)
- Fixes some bindings to Document APIs

### DIFF
--- a/test/dune
+++ b/test/dune
@@ -1,13 +1,8 @@
 (melange.emit
- (alias jest)
+ (alias runtest)
  (target test)
  (module_systems
   (commonjs bs.js))
  (libraries reason-react reason-react.node jest melange.belt)
  (preprocess
   (pps melange.ppx reason-react-ppx)))
-
-(rule
- (alias runtest)
- (deps (alias_rec jest))
- (action (run npx jest)))


jchavarri:
Oh, just realized we were running jest explicitly in CI. Happy to roll that part back (I think I get confused because I expect `runtest` to do everything for me...)

src/ReactDOMTestUtils.re
```diff
@@ -84,7 +84,7 @@ module Simulate = {
 
 external document: Dom.document = "document";
 
-[@mel.send]
+[@mel.return nullable] [@mel.send]
```
anmonteiro, line 87: for my understanding: we're adding `@mel.return nullable` because `querySelector` may return `null` or `undefined` and we want to convert it to an option?

anmonteiro (approved):

src/ReactDOMTestUtils.re
jchavarri: Actually it can only return `null`. https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector

Maybe I should use `null_to_opt`? To be more specific.

anmonteiro: this might be a good idea, but no strong opinions

PR #830, opened 2024-03-27
Make explicit array comparison error in useEffectN N>1

benbellick:
I'm working on a project using reason-react and I encountered a mistake which I found hard to solve, and I wanted to contribute to the docs in case anyone else experiences the same issue. 

The gist is that my reading of the docs led me to believe that the use of arrays in `useEffect` are fine in general, rather than just specifically for `useEffect1`. 

So, I tried to be clever by gathering like types into arrays, like so:
```reason
x : int
y : int
z : string
useEffect2(effect, [|x,y|],z);
```

But this doesn't work, as the transpiled javascript code will always result in a comparison `[x,y]==[x,y]` which is `false` always because the arrays are being compared by reference. 

I'm not sure if this is the right place to plop these docs in, but please let me know and I am happy to move it around!

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -109,6 +109,18 @@ useEffect1(effect, [|dep|])
 
 However, as the length of the array is not specified, you could pass an array of arbitrary length, including the empty array, `[||]`.
 
+> *NOTE*: When using `useEffectN` for `N` greater than 1, do not try to gather like types into arrays. Instead, leave each item separate in the tuple. 
+> As an example, consider the case where your effect has three deps, with the first and second dep the same type. The reason call:
+> ```reason
+> useEffect1(effect, ([|dep1, dep2|], dep3))
+> ```
+> would be expressed in javascript as:
+> ```javascript
+> useEffect(effect, [[dep1, dep2], dep3])
+> ```
+> Because arrays are compared by reference, this `useEffect` call will identify an update in its dependency array every time. 
+
+
 Reason also always opts for the safest form of a given hook as well. So `React.useState` in JS can take an initial value or a function that returns an initial value. The former cannot be used safely in all situations, so ReasonReact only supports the second form which takes a function and uses the return.
 
 ## Hand-writing components
--- a/docs/components.md
+++ b/docs/components.md
@@ -109,17 +109,10 @@ useEffect1(effect, [|dep|])
 
 However, as the length of the array is not specified, you could pass an array of arbitrary length, including the empty array, `[||]`.
 
-> *NOTE*: When using `useEffectN` for `N` greater than 1, do not try to gather like types into arrays. Instead, leave each item separate in the tuple. 
-> As an example, consider the case where your effect has three deps, with the first and second dep the same type. The reason call:
-> ```reason
-> useEffect2(effect, ([|dep1, dep2|], dep3))
-> ```
-> would be expressed in javascript as:
-> ```javascript
-> useEffect(effect, [[dep1, dep2], dep3])
-> ```
-> Because arrays are compared by reference, this `useEffect` call will identify an update in its dependency array every time. 
-
+> *NOTE*: When using `useEffectN` for `N` greater than 1, be careful not try to gather like types into arrays. Instead, leave each item separate in the tuple.
+> This is because the compiled javascript will compare by object, and thus will always interpret different arrays as unequal, regardless of their content.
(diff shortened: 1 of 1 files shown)


docs/components.md
```diff
@@ -109,6 +109,18 @@ useEffect1(effect, [|dep|])
 
 However, as the length of the array is not specified, you could pass an array of arbitrary length, including the empty array, `[||]`.
 
+> *NOTE*: When using `useEffectN` for `N` greater than 1, do not try to gather like types into arrays. Instead, leave each item separate in the tuple.
```
davesnx, line 112: I would explain this a bit differently:

> Using useEffectN1/N2/N3/... make sure you don't wrap the dependencies in arrays. Since tuples are represented in arrays in JavaScript it would trick React into compare arrays by reference (which is often undesired).

> Use the tuple given one dependency per item, even when those types are the same.

anmonteiro:
I'd probably rather just point to the React docs where this is explained:

From https://react.dev/reference/react/useEffect#reference :

> If some of your dependencies are objects or functions defined inside the component, there is a risk that they will cause the Effect to re-run more often than needed. To fix this, remove unnecessary [object](https://react.dev/reference/react/useEffect#removing-unnecessary-object-dependencies) and [function](https://react.dev/reference/react/useEffect#removing-unnecessary-function-dependencies) dependencies. You can also [extract state updates](https://react.dev/reference/react/useEffect#updating-state-based-on-previous-state-from-an-effect) and [non-reactive logic](https://react.dev/reference/react/useEffect#reading-the-latest-props-and-state-from-an-effect) outside of your Effect.


From https://react.dev/learn/synchronizing-with-effects#re-render-with-different-dependencies :

> The dependency array can contain multiple dependencies. React will only skip re-running the Effect if all of the dependencies you specify have exactly the same values as they had during the previous render. React compares the dependency values using the [Object.is](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) comparison. See the [useEffect reference](https://react.dev/reference/react/useEffect#reference) for details.

benbellick:
How's the change now? I'm happy to remove the extra stuff if you feel strongly. My view is that the type- ystem may incorrectly guide the user to think the `useEffectN` applied on an array is working because it compiles. This was a particular place I got stuck on which took some time to fix, and so I would rather include some info on the specific issue directly in the docs so as to make easiest for users.

actionshrimp:
Having been confused by this in the past in the past, I think the issue stems from the progression from the singleton array -> int or string array -> int or string array + some other type. What about something [like this](https://github.com/reasonml/reason-react/commit/a8ec4b34b1b0c1e84a2ccbf1028bb83df29dece2), including the progression in the example compiled output?

```reason
useEffect1(effect, [| 1 |])
     /* ^^^ -- Compiles to javascript as `useEffect(effect, [1])` */
useEffect1(effect, [| 1, 2, 3 |])
     /* ^^^ -- Compiles to javascript as `useEffect(effect, [1, 2, 3])` */
useEffect2(effect, [| 1, 2, 3 |], "foo")
     /* ^^^ --- !! Compiles to javascript as `useEffect(effect, [ [1, 2, 3], "foo" ])` */
```

davesnx:
I'm confused now `useEffect2(effect, [| 1, 2, 3 |], "foo")` isn't valid to compile, which is the point: https://melange.re/unstable/playground/?language=Reason&code=UmVhY3QudXNlRWZmZWN0MihlZmZlY3QsIFt8MSwgMiwgM3xdLCAiZm9vIik7Cg%3D%3D&live=on

actionshrimp:
My bad, I was typing this up in a bit of a rush and missed adding the tuple wrapper for `useEffect2` which I've now edited in above. 

I think that mistake partly explains my initial confusion with this stuff though when I started with ReasonReact - I don't really think of tuples and arrays in the same way intuitively, even if they both compile to an array, so I hadn't appreciated that the API was using the tuple in place of the array to try and mimick how things look in javascript - I just saw the 2, and saw that you passed it 2 things.

So following that logic, `useEffect1` always seemed like the odd one out to me:
- `useEffect0` takes 0 things
- `useEffect1` takes an array
- `useEffect2` takes 2 things
- `useEffect3` takes 3 things etc..

I guess ideally it might look something like this, if this were possible:

```
useEffect0(effect)
useEffect1(effect, (1)) <-- unfortunately this isn't valid either
useEffect2(effect, (1, 2))
...
useEffectA(effect, [|1, 2, ...|])
```

davesnx:
I would open a PR explaining https://github.com/reasonml/reason-react/pull/830#issuecomment-2047004023 on the documentation, which is a bit clearer and mention your point @benbellick 

Thanks

Issue #834, opened 2024-06-03
Fragment + new jsx transform


tatchi:
Given the following code

```js
const App = () => <>hello</>
```

And using the new `jsx` transform, babel compiles it down to ([link](https://babeljs.io/repl#?browsers=%3E1%25%2C%20last%204%20versions%2C%20Firefox%20ESR%2C%20not%20dead&build=&builtIns=false&corejs=false&spec=false&loose=false&code_lz=MYewdgzgLgBAggBwTAvDAFASlQPhgHhwAsBTAGzJHwHocAoBoA&debug=false&forceAllTransforms=true&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=react%2Ctypescript&prettier=false&targets=&version=7.24.6&externalPlugins=&assumptions=%7B%7D))

```js
import { Fragment as _Fragment, jsx as _jsx } from "react/jsx-runtime";
const App = () => /*#__PURE__*/_jsx(_Fragment, {
  children: "hello"
});
```

While `reason-react` still uses `React.createElement` ([playground](https://melange.re/v4.0.0/playground/?language=Reason&code=W0ByZWFjdC5jb21wb25lbnRdCmxldCBtYWtlID0gKCkgPT4gPD57UmVhY3Quc3RyaW5nKCJoZWxsbyIpfTwvPgo%3D&live=off))

```js
// Generated by Melange

import * as React from "react";
import * as JsxRuntime from "react/jsx-runtime";

function _none_(Props) {
  return React.createElement(JsxRuntime.Fragment, undefined, "hello");
}

const make = _none_;

export {
  make ,
}
/* react Not a pure module */

```

jchavarri:
I think this issue can be closed now (thanks @tatchi !)

PR #835, opened 2024-06-03
support jsx transform with fragment

tatchi:
This is my attempt at fixing https://github.com/reasonml/reason-react/issues/834

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+* Support JSX transform with fragments (@tatchi in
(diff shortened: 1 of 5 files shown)


anmonteiro:
Thank you! This is a great catch.

I was wondering whether `< key=... > </>` was possible but it looks like it isn't: https://react.dev/reference/react/Fragment#caveats

anmonteiro:
I fixed the transformation. in the jsx runtime, children need to be passed in the props object as `{children: ...}`

PR #833, opened 2024-04-25
melange v4 long-lived branch

anmonteiro:

PR #836, opened 2024-06-03
require melange 4 and udpate tests

tatchi:
I had melange v4 installed when I did `make init`, which produces some diffs in tests. This is due to the `var` -> `const` change in melange v4. This PR suggests bumping the required melange version to avoid this.

I'm not a `nix` user, so I don't know if anything needs to be updated on that side.


jchavarri:
Please check https://github.com/reasonml/reason-react/pull/833.

tatchi:
My bad, I'm closing this one then

PR #838, opened 2024-06-07
feat(react): add startTransition

r17x:
Reference: https://react.dev/reference/react/startTransition

### DIFF
--- a/src/React.re
+++ b/src/React.re
@@ -866,6 +866,9 @@ module Uncurried = {
     "useCallback";
 };
 
+[@mel.module "react"]
+external startTransition: ([@mel.uncurry] (unit => unit)) => unit = "startTransition";
+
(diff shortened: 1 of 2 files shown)


davesnx:
Thanks @r17x

src/React.re
```diff
@@ -872,10 +872,15 @@ module Uncurried = {
     "useCallback";
 };
 
+type callback('input, 'output) = 'input => 'output;
```
anmonteiro, line 875: this is already defined in line 471

anmonteiro: just saw https://github.com/reasonml/reason-react/commit/757dd23e1b1c22a2fc82fd48b4951c65e84235d2

r17x: Sorry for confusing - This PR should be PASSED Pipelines before merge. @anmonteiro

PR #839, opened 2024-06-10
refactor: convert error boundary code to Reason

anmonteiro:
generated code looks like:

```js
// Generated by Melange
'use strict';

const Js__Js_undefined = require("melange.js/js_undefined.bs.js");
const React = require("react");

function errorBoundary(_props) {
  let self = this ;
  React.Component.call(self);
  self.state = {
    error: undefined
  };
}

errorBoundary.prototype = Object.create(React.Component.prototype);

errorBoundary.prototype.componentDidCatch = (function (error, info) {
    let self = this ;
    self.setState({
          error: {
            error: error,
            info: info
          }
        });
  });

errorBoundary.prototype.render = (function () {
    let self = this ;
    if (Js__Js_undefined.testAny(self.state.error)) {
      return self.props.children;
    } else {
      return self.props.fallback(self.state.error);
    }
  });

const make = errorBoundary;

exports.make = make;
/* errorBoundary Not a pure module */
```

### DIFF
--- a/src/ReasonReactErrorBoundary.re
+++ b/src/ReasonReactErrorBoundary.re
@@ -11,32 +11,62 @@ type params('error) = {
   info,
 };
 
-[%%mel.raw
-  {|
-  var React = require("react");
-  var ErrorBoundary = (function (Component) {
-    function ErrorBoundary(props) {
-      Component.call(this);
-      this.state = {error: undefined};
-    };
-    ErrorBoundary.prototype = Object.create(Component.prototype);
-    ErrorBoundary.prototype.componentDidCatch = function(error, info) {
-      this.setState({error: {error: error, info: info}});
-    };
-    ErrorBoundary.prototype.render = function() {
-      return this.state.error != undefined
-        ? this.props.fallback(this.state.error)
-        : this.props.children
-    };
-    return ErrorBoundary;
-  })(React.Component);
-|}
-];
+[@mel.scope "Object"] external objCreate: 'a => Js.t({..}) = "create";
+
+[@ocaml.warning "-69"]
+type state('error) = {error: params('error)};
+type reactComponentClass;
+
+[@mel.module "react"] external component: reactComponentClass = "Component";
+[@mel.send] external componentCall: (reactComponentClass, _) => unit = "call";
+
+type componentPrototype;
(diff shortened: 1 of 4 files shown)


src/ReasonReactErrorBoundary.re
```diff
@@ -11,32 +11,60 @@ type params('error) = {
...
-    ErrorBoundary.prototype.render = function() {
-      return this.state.error != undefined
-        ? this.props.fallback(this.state.error)
-        : this.props.children
-    };
-    return ErrorBoundary;
-  })(React.Component);
-|}
-];
+[@mel.scope "Object"] external objCreate: 'a => Js.t({..}) = "create";
```
davesnx, line 14: If you want to add `melange.js`, can use Obj.empty

davesnx (approved):
Very cursed code, but nicely done

src/ReasonReactErrorBoundary.re
```diff
@@ -11,32 +11,60 @@ type params('error) = {
...
+
+[@mel.module "react"] external component: reactComponentClass = "Component";
+[@mel.send] external componentCall: (reactComponentClass, _) => unit = "call";
+
+type componentPrototype;
+[@mel.get]
+external componentPrototype: reactComponentClass => componentPrototype =
+  "prototype";
+
+[@ocaml.warning "-21"]
```
davesnx, line 28: I trend to add the why when I disable some warnings, since 21/69 isn't descriptive enough

src/ReasonReactErrorBoundary.re
anmonteiro: `Js.Obj.empty` doesn't allow you to do `Object.create(React.Component.prototype)`.

src/ReasonReactErrorBoundary.re
anmonteiro: good idea! I realized we don't actually need these at all. check a8d20ef2c6acd2334a5597b705402bef6c14eb9f

src/ReasonReactErrorBoundary.re
davesnx: Ops, obviusly. Could add obj.create on `melange.js`

Issue #840, opened 2024-06-12
Type-on-hover stops working inside components with more than one prop on OCaml 5.2


feihong:
Here's a small example:

```reason
[@react.component]
let make = (~foo, ~bar) => {
  <div> {React.string(foo)} {bar |> string_of_int |> React.string} </div>;
};
```
In VS Code, if you hover over `foo` or `bar`, you get `(~foo: string, ~bar: int) => React.element`, which is the type signature of the function. There are at least two ways to get normal type-on-hover back:

- Remove one of the props, either `foo` or `bar`
- Remove the `[@react.component]` attribute

Note that the problem isn't limited to Reason syntax, it also doesn't work with OCaml syntax:

```ocaml
let make ~foo  ~bar  =
  ((div ~children:[React.string foo; (bar |> string_of_int) |> React.string]
      ())
  [@JSX ]) [@@react.component ]
```

I created a small repo that demonstrates the problem: https://github.com/feihong/type-on-hover-bug

PR #842, opened 2024-06-23
test: repro #840

anmonteiro:


ppx/test/hover.t
```diff
@@ -0,0 +1,76 @@
...
+    {
+      "start": {
+        "line": 1,
+        "col": 0
+      },
+      "end": {
+        "line": 4,
+        "col": 15
+      },
+      "type": "< bar : int; foo : string > Js.t -> React.element",
```
anmonteiro, line 73: this reproduces the issue @feihong reported: this type should be `string`, but instead it's the entire function type.

jchavarri (approved):
Only tangentially related. This PR made me wonder if instead of disabling the failing test like we did in https://github.com/reasonml/reason-react/pull/833/commits/14ac9e0acba6bfb7193394e0a9b9525d9bdfdcc7 we should rather update it to reflect reality, for completion.

PR #844, opened 2024-07-01
Add locations-check test

jchavarri:
I found out today that ppxlib has some `-check -locations-check` flag that one can use to make sure a ppx generates an AST with valid locations ("valid" in the sense of following the Merlin rules).

I thought this could serve as an alternative to tests like the one @anmonteiro adds in #842 and the many other merlin-based tests that we have.

Rather than building the test based on what merlin returns, we can just make sure the tree is well formed. But it turns out the check doesn't successfully pass (even in 5.1), because there are too many overlaps.

Maybe we can keep the test as is for now and over time solve all issues until we make sure it passed this check. @davesnx @anmonteiro wdyt?

### DIFF
--- a/ppx/test/locations-check.t
+++ b/ppx/test/locations-check.t
@@ -1,0 +1,66 @@
+Test the preprocessed reason-react components have well-formed locations.
+Uses https://github.com/ocaml-ppx/ppxlib/blob/44583fc14c3cc39ee6269ffd69f52146283f72c0/src/location_check.mli
+
+With no annotations (ppx does nothing)
+
+  $ cat >input.ml <<EOF
+  > let make ~foo ~bar =
+  >   (div
+  >      ~children:[ React.string foo; bar |> string_of_int |> React.string ]
+  >      ())
+  > EOF
+
+  $ reason-react-ppx -check -locations-check input.ml
+  let make ~foo  ~bar  =
+    div ~children:[React.string foo; (bar |> string_of_int) |> React.string] ()
+
+With JSX annotation
+
+  $ cat >input.ml <<EOF
+  > let make ~foo ~bar =
+  >   (div
+  >      ~children:[ React.string foo; bar |> string_of_int |> React.string ]
+  >      () [@JSX])
+  > EOF
+
+  $ reason-react-ppx -check -locations-check input.ml
+  File "input.ml", line 2, characters 3-6:
+  2 |   (div
+         ^^^
+  Error: invalid output from ppx, expression overlaps with expression at location:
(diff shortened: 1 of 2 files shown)


ppx/test/locations-check.t
```diff
@@ -0,0 +1,66 @@
...
+  >   (div
+  >      ~children:[ React.string foo; bar |> string_of_int |> React.string ]
+  >      () [@JSX])
+  > EOF
+
+  $ reason-react-ppx -check -locations-check input.ml
+  File "input.ml", line 1, characters 33-36:
+  1 | let[@react.component] make ~foo ~bar =
+                                       ^^^
+  Error: invalid output from ppx, core type overlaps with core type at location:
```
davesnx, line 47: Woa

davesnx (approved):

anmonteiro (approved):
pretty nice. 

I wonder if the failing tests will now pass after the ppxlib 0.33 release

jchavarri:
> I wonder if the failing tests will now pass after the ppxlib 0.33 release

They do. I just had to modify one location in the ppx tests (seems not harmful).

jchavarri:
> They do. I just had to modify one location in the ppx tests (seems not harmful).

Hm just realized this change was already in https://github.com/reasonml/reason-react/pull/850. I forgot to merge `main` into this branch 🤦

ppx/test/locations-check.t
```diff
@@ -0,0 +1,66 @@
...
+  $ cat >input.ml <<EOF
+  > let make ~foo ~bar =
+  >   (div
+  >      ~children:[ React.string foo; bar |> string_of_int |> React.string ]
+  >      () [@JSX])
+  > EOF
+
+  $ reason-react-ppx -check -locations-check input.ml
+  File "input.ml", line 2, characters 3-6:
+  2 |   (div
```
anmonteiro, line 28: is this expected? What exactly is this test testing if these are errors?

anmonteiro: or are these things we should put on our TODO list to fix?

jchavarri: Yes, this should go in a TODO list to fix. The idea behind it is that if the ppx creates the AST in a way that this tests pass, we should never have any more problems with editor not finding locations, getting confused about types on hover, etc, as the tree would be consistent.

jchavarri: Rather than having to write "manual" tests in a per-case basis like in https://github.com/reasonml/reason-react/pull/842.

PR #846, opened 2024-07-16
Support for React 19

davesnx:
This PR adds some of the new functions on React 19.

- `useActionState` It relies on FormData and depends on melange.dom's PR (https://github.com/melange-re/melange/pull/1153, for passing the build and testing other stuff FormData is vendored which we won't publish with FormData here).
- Binds to **`useOptimistic`**
- Change `Experimental.use` to `Experimental.usePromise` and add `Experimental.useContext` 
- Change `act` in react-dom/test-utils
- Add deprecation in ReactDOMTestUtils
- Merged https://github.com/reasonml/reason-react/pull/862 into it
- Merged https://github.com/reasonml/reason-react/pull/859 into it


src/FormData.ml
```diff
@@ -0,0 +1,53 @@
+(* Provides a way to easily construct a set of key/value pairs representing form fields and their values, which can then be easily sent using the XMLHttpRequest.send() method. It uses the same format a form would use if the encoding type were set to "multipart/form-data". *)
+
+module Iterator = struct
```
anmonteiro, line 3: I think we added iterator stuff in melange, can you reuse that?

src/ReactDOM.re
```diff
@@ -648,7 +648,7 @@ type domProps = {
   [@mel.optional]
   acceptCharset: option(string),
   [@mel.optional]
-  action: option(string), /* uri */
+  action: option(FormData.t => Js.Promise.t(unit)), /* Since action is taken by "form" as string and React 19 accepts a callback we keep a 'action_' field to avoid a breaking change. */
```
anmonteiro, line 651: what do you mean by "keep a 'action_'" field? isn't the field still named `action` (without underscore?)

src/FormData.ml
```diff
@@ -0,0 +1,53 @@
...
+  external toArray : 'a t -> 'a array = "Array.from"
+  external toArrayWithMapper : 'a t -> f:('a -> 'b) -> 'b array = "Array.from"
+end
+
+type t
+type file
+type blob
+type entryValue
+
+let classify : entryValue -> [> `String of string | `File of file ] =
```
anmonteiro, line 18: I think we could just use `Js.Types.classify` as I mentioned in the melange PR comment.

src/React.re
```diff
@@ -886,6 +887,31 @@ external useDebugValue: ('value, ~format: 'value => string=?, unit) => unit =
 
 module Experimental = {
   /* This module is used to bind to APIs for future versions of React. There is no guarantee of backwards compatibility or stability. */
+  [@mel.module "react"] external usePromise: Js.Promise.t('a) => 'a = "use";
+  [@mel.module "react"] external useContext: Context.t('a) => 'a = "use";
+  [@mel.module "react"] external use: 'a => 'b = "use";
```
anmonteiro, line 892: should we default the name `use` to `Js.Promise.t('a)` and add `useAny` for `'a => 'b`?

src/React.re
```diff
@@ -886,6 +887,31 @@ external useDebugValue: ('value, ~format: 'value => string=?, unit) => unit =
...
+  [@mel.module "react"]
+  external useTransitionAsync:
+    unit => (bool, callbackAsync(callbackAsync(unit, unit), unit)) =
+    "useTransition";
+
+  /* https://es.react.dev/reference/react/useOptimistic */
+  [@mel.module "react"]
+  external useOptimistic:
+    ('state, ('state, 'optimisticValue) => 'state) =>
+    ('state, 'optimisticValue => unit) =
```
anmonteiro, line 903: > addOptimistic is the dispatching function to call when you have an optimistic update. It takes one argument, optimisticValue, of any type and will call the updateFn with state and optimisticValue.

if it'll call the `updateFn` it will return `'state`

```suggestion
    ('state, 'optimisticValue => 'state) =
```

anmonteiro: nevermind, definitely typed says it returns `void`:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/136deed9a5c0adaea7bc7674783bb52608673d05/types/react/ts5.0/canary.d.ts#L91-L97

src/ReactDOM.re
anmonteiro: I believe this is now a breaking change because we don't accept a string for `action` anymore

src/React.re
andreypopp: What's the use case for `use : 'a -> 'b`, which is essentially `Obj.magic`?

src/ReactDOM.re
davesnx: I couldn't find a way to keep both (you can't have 2 mel.as with the same name), so it's either we don't support action like this (and users could cloneElement it) or we break with action.

Not fan of either

andreypopp: as I understand `action` is only for `<form>`? maybe create `<Form />`, a separate component specifically for RSC apps, not to make API more complex for a majority of use cases?

anmonteiro: we thought we'd be able to support this by e.g. adding `@mel.unwrap` support in `@@deriving jsProperties` but it turns out it's unsound (some discussion in https://github.com/melange-re/melange/pull/1162).

it'd be kinda nice if ppxlib allowed to make the original type abstract, which would remove the unsoundness.

src/React.re
```diff
@@ -886,6 +887,71 @@ external useDebugValue: ('value, ~format: 'value => string=?, unit) => unit =
...
 module Experimental = {
   /* This module is used to bind to APIs for future versions of React. There is no guarantee of backwards compatibility or stability. */
+  [@mel.module "react"] external usePromise: Js.Promise.t('a) => 'a = "use";
+  [@mel.module "react"] external useContext: Context.t('a) => 'a = "use";
+  [@mel.module "react"]
+  external useTransitionAsync:
+    unit => (bool, callbackAsync(callbackAsync(unit, unit), unit)) =
+    "useTransition";
+
+  /* https://es.react.dev/reference/react/useOptimistic */
```
jchavarri, line 897:
```suggestion
  /* https://react.dev/reference/react/useOptimistic */
```

src/React.re
```diff
@@ -886,6 +887,71 @@ external useDebugValue: ('value, ~format: 'value => string=?, unit) => unit =
 
 module Experimental = {
   /* This module is used to bind to APIs for future versions of React. There is no guarantee of backwards compatibility or stability. */
+  [@mel.module "react"] external usePromise: Js.Promise.t('a) => 'a = "use";
+  [@mel.module "react"] external useContext: Context.t('a) => 'a = "use";
+  [@mel.module "react"]
```
jchavarri, line 892:
```suggestion
  /* https://react.dev/reference/react/useTransition */
  [@mel.module "react"]
```

src/React.rei
```diff
@@ -565,15 +566,81 @@ module Uncurried: {
...
+  [@mel.module "react"]
+  external useOptimistic:
+    ('state, ('state, 'optimisticValue) => 'state) =>
+    ('state, 'optimisticValue => unit) =
+    "useOptimistic";
+
+  [@mel.module "react"]
+  external useTransitionAsync:
+    unit => (bool, callbackAsync(callbackAsync(unit, unit), unit)) =
+    "useTransition";
```
jchavarri, line 590: Prob out of scope, but now that new bindings are being added, and given they're new and experimental, it's good chance to use chatgpt to add comments to them? i got this for this one with some tweaking:

```suggestion
  /**
   * `useTransitionAsync` is a binding to React's `useTransition` hook.
   * 
   * This hook lets you update state without blocking the UI.
   * 
   * ### Return Value
   * - `(bool, callbackAsync(callbackAsync(unit, unit), unit))`: A tuple containing:
   *   - `bool`: Whether the transition is pending.
   *   - A function to start the transition.
   * 
   * ### Example
   * ```reason
   * let (isPending, startTransition) = useTransitionAsync();
   *
   * /* Start a transition */
   * startTransition(() => {
   *   /* Deferred update logic */
   * });
   * 
   * if (isPending) {
   *   /* Show loading indicator */
   * }
   * ```
   */
  [@mel.module "react"]
  external useTransitionAsync:
    unit => (bool, callbackAsync(callbackAsync(unit, unit), unit)) =
    "useTransition";
```

src/ReactDOM.re
```diff
@@ -652,7 +652,7 @@ type domProps = {
   [@mel.optional]
   acceptCharset: option(string),
   [@mel.optional]
-  action: option(string), /* uri */
```
jchavarri, line 655: This comment and the one in `rei` were useful imo. Or were they not accurate?

src/ReactDOMTestUtils.re
```diff
@@ -2,7 +2,7 @@ type undefined = Js.undefined(unit);
 
 let undefined: undefined = Js.Undefined.empty;
 
-[@mel.module "react-dom/test-utils"]
+[@mel.module "react"]
```
jchavarri, line 5: If this is not gonna be in `react-dom` anymore, I find it confusing to keep it in `ReactDOMTestUtils`. Should we move it to `React` for consistency / principle of less surprise?

davesnx: Good point, I should probably keep both and add a deprecation

jchavarri: idk if deprecation is necessary. A ton of other things are breaking upstream for React 19, so maybe we can just move it directly.

src/React.re
```diff
@@ -886,6 +887,73 @@ external useDebugValue: ('value, ~format: 'value => string=?, unit) => unit =
...
+  [@mel.module "react"] external useContext: Context.t('a) => 'a = "use";
+  /* https://react.dev/reference/react/useTransition */
+  [@mel.module "react"]
+  external useTransitionAsync:
+    unit => (bool, callbackAsync(callbackAsync(unit, unit), unit)) =
+    "useTransition";
+
+  /* https://react.dev/reference/react/useOptimistic */
+  [@mel.module "react"]
+  external useOptimistic:
```
anmonteiro, line 901: can we `@mel.uncurry` some of these callbacks?

src/React.re
```diff
@@ -886,6 +887,73 @@ external useDebugValue: ('value, ~format: 'value => string=?, unit) => unit =
...
+    external entries: t => Js.Iterator.t((string, entryValue)) = "entries";
+  };
 
-  [@mel.module "react"] external use: Js.Promise.t('a) => 'a = "use";
+  type formAction;
+
+  /* https://react.dev/reference/react/useActionState */
+  [@mel.module "react"]
+  external useActionState:
+    (('state, FormData.t) => unit, 'state, ~permalink: bool=?) =>
```
anmonteiro, line 956: should we add a unit argument at the end?

src/React.rei
anmonteiro: I think this is a good idea but I'd recommend doing it in a future PR. this is already quite a bit dense for revewing.

src/ReactDOM.re
anmonteiro: @davesnx wasn't there something about this field that bothered you a bit? 

can you link us to the issue about this? I think something about `action` being a function for RSC or something?

src/ReactDOMTestUtils.re
anmonteiro: agree this should be moved to react.re if it's coming from `require('react')`

anmonteiro:
@davesnx this is a big PR and there are some big changes. do you have a plan in your head or a link to something you've written about how to land these effectively in reason-react? 

more specifically about this PR, how do you want to move forward and what's missing?

davesnx:
I was planning on using the branch "19" as a release candidate branch for all PRs depending on it, test it different repos (ahrefs included) and, if everything is good then do the opam release

src/ReactDOM.re
davesnx: Reverting this commit.

`action` is problematic in React 19 since it overloads it. formAction and action in this PR they are sync (while in React can be async and have different behaviour), if users want to use it they would need to use createElement + cloneElement or one of those tricks.

src/React.rei
```diff
@@ -565,15 +566,81 @@ module Uncurried: {
...
+  [@mel.module "react"] external useContext: Context.t('a) => 'a = "use";
+
+  [@mel.module "react"]
+  external useOptimistic:
+    ('state, ('state, 'optimisticValue) => 'state) =>
+    ('state, 'optimisticValue => unit) =
+    "useOptimistic";
+
+  [@mel.module "react"]
+  external useTransitionAsync:
```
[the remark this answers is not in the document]
davesnx: `useTransition` is sync (and stable since R18) and this is `Experimental.useTransitionAsync`

src/ReactDOMTestUtils.re
davesnx: There are a few "react" in ReactDOM module. I might probably propose to have this as a strict rule and align those as well in a future PR

src/React.re
davesnx: Since I don't want to push code to this branch directly, done here: https://github.com/reasonml/reason-react/pull/867

davesnx:
This PR is waiting to be merged for a release from react.js @ 19.1.0 with the fix required to work, currently pointed at `react@19.1.0-canary-662957cc-20250221`

davesnx:
React 19, here we go!

PR #852, opened 2024-08-15
Fix multi-child fragment

jchavarri:
Currently, this code `<> <div/> <div/> </>` will trigger the infamous warning: `Each child in a list should have a unique "key" prop`.

This PR modifies the expression switch for fragments so that those with 0 or 1 child use `jsx` and those with 2 or more children use `jsxs`.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,1 +1,1 @@
-version = 0.26.0
+version = 0.26.2
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,1 +1,1 @@
-version = 0.26.2
+version = 0.26.0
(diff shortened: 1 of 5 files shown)


davesnx (approved):
Nice fix

.ocamlformat
```diff
@@ -1 +1 @@
-version = 0.26.0
+version = 0.26.2
```
tatchi, line 1: Is this change on purpose or because you had this version locally in your switch? IOW, should it be commited?

tatchi:
Nice, thanks! 🤗

.ocamlformat
jchavarri: You're right, reverted in https://github.com/reasonml/reason-react/pull/852/commits/e5eeb3f23fd08fe028f7220f87caa5b86d3161c2.

anmonteiro (approved):
I pushed a small refactor that avoids sharing the code in the entire branch (just decides between `jsx` and `jsxs`)

PR #859, opened 2024-11-19
Replace react dom's testing library with ReactTestingLibrary

davesnx:
Depends on #846 

- Vendors ReactTestingLibrary (and does some ad-hoc changes) because React 19 [deprecates react-dom test utils](https://react.dev/warnings/react-dom-test-utils#rest-of-reactdomtestutils-apis).
- Creates a `demo` folder to ensure React works manually, loads, etc.
- Migrates all jest tests to use RTL
- Currently uses React.Uncurried for useState and useReducer waiting for https://github.com/reasonml/reason-react/pull/860 and to have a better solution from https://github.com/facebook/react/issues/31578

### DIFF
--- a/test/Form__test.re
+++ b/test/Form__test.re
@@ -136,7 +136,7 @@ describe("Form with useOptimistic", () => {
     let container = ReactTestingLibrary.render(<App />);
 
     ReactTestingLibrary.actAsync(() => {
-      let.await _ = findByString("Hola!", container);
+      let.await _ = findByString({j|¡Hola!|j}, container);
 
       let.await button = findByString("Enviar", container);
       let.await input = findByPlaceholderText("message", container);


test/melange-testing-library/README.md
```diff
@@ -0,0 +1,5 @@
+### vendored melange-testing-library
```
jchavarri, line 1: I read the PR description but it explains what's being done but not why.

What was the issue with using react test renderer? It's a much lighter solution than react testing library and it should catch any runtime issues just fine.

test/melange-testing-library/README.md
```diff
@@ -0,0 +1,5 @@
...
+
+This directory contains a vendored version of the melange-testing-library.
+
+The original repository is https://github.com/melange-community/melange-testing-library
```
jchavarri, line 5: The PR description mentions

> Vendors ReactTestingLibrary (and does some ad-hoc changes)

are the ad-hoc changes valuable for other users of the bindings? in that case, will they be upstreamed?

jchavarri (commented):
I'm afraid I don't know the reason why switching to another (heavier) testing library is needed.

test/melange-testing-library/README.md
davesnx: Not sure, melange-testing-library contains a few binding designs focused on previous versions of BuckleScript where options aren't translated to undefined, or records aren't objects.

I would keep the changes here and evolve this bindings as a playground for the melange-testing-library

test/melange-testing-library/README.md
jchavarri: ftr: https://react.dev/warnings/react-dom-test-utils#rest-of-reactdomtestutils-apis

Makefile
```diff
@@ -54,11 +59,19 @@ format-check: ## Checks if format is correct
 .PHONY: install
 install: ## Update the package dependencies when new deps are added to dune-project
 	@opam install . --deps-only --with-test
-	@npm install
+	@npm install --force
```
jchavarri, line 62: Why was `force` needed?

demo/dune
```diff
@@ -0,0 +1,9 @@
+(melange.emit
+ (target demo)
+ (alias melange-app)
+ (module_systems
+  (es6 mjs))
+ (libraries reason-react jest melange.belt melange.dom)
```
jchavarri, line 6: Why is `jest` needed for the demo?

jchavarri (approved):

demo/dune
davesnx: Not needed indeed

Makefile
davesnx: Because testinglib and other deps depend on react 18, so npm complains. It is a decent workaround to ensure the installation succeeds

davesnx: Will write a comment

test/melange-testing-library/README.md
anmonteiro: works, but we should upstream as they get better!

Makefile
anmonteiro: would we still needed `--force` if you commit the lockfile from manual install and run `npm ci` here?

anmonteiro (approved):

PR #862, opened 2024-11-20
Enable ref as valid prop

davesnx:
Depends on #859, that depends on #846 

Allows to have a prop with `~ref`

---

This PR also adds the cleanup function on `ref` as a new constructor: added another "constructor" instead of changing the current callbackRef to avoid a breaking change, we might add deprecations and further cleanup in the future versions to keep those constructors to the minimum


test/Ref__test.re
```diff
@@ -0,0 +1,29 @@
...
+  test("can be passed as prop to a component", () => {
+    let domRef = React.createRef();
+    let container =
+      ReactTestingLibrary.render(
+        <Button_with_ref ref={ReactDOM.Ref.domRef(domRef)}>
+          {React.string("Click me")}
+        </Button_with_ref>,
+      );
+    let button = getByRole("FancyButton", container);
+    expect(button->innerHTML)->toBe("Click me");
```
jchavarri, line 27: Shouldn't we test here something about the `ref`? This test just shows that `Button_with_ref` is rendering ok.

jchavarri (approved):

test/Ref__test.re
davesnx: Sure, added another assertion but didn't want to go to clicks and other events to keep the test simple and fast

jchavarri: Thanks. (I think the expect in line 27 can be removed, because it doesn't test anything related to the ref)

anmonteiro:
context: https://vercel.com/blog/whats-new-in-react-19#ref-as-a-prop

Issue #843, opened 2024-06-27
Very misleading error message if you misuse [@react.component]


Josef-Thorne-A:
If you put [@reason.component] in the wrong location it will give a very confusing and uninformative error. This took me a couple hours to track down properly:

Example of problematic code:
```
[@react.component]`
let (pageState,setPageState) = React.useState(_ => testState);

let make { ...}
```

For me at least in an error message in the associated dune file was given, nowhere near where the issue was -- nor obviously related to incorrect ppx directives:

```
7 |  (preprocess (pps melange.ppx reason-react-ppx)))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Fatal error: exception Invalid_argument("react.component calls cannot be destructured."
```

PR #863, opened 2024-11-20
Add test for error on key

davesnx:
Depends on #862 

Minimalist and spiritual successor of https://github.com/reasonml/reason-react/pull/848

Fix https://github.com/reasonml/reason-react/issues/843

Took the chance to cleanup all `raise Invalid_argument` into `Location.raise_errorf` (which isn't ideal from the ppxlib standpoint, since it fails a Parseetree generation and only shows one error) but the refactor to this is massive


davesnx:
Made it depend on #main @jchavarri

PR #880, opened 2025-02-24
Revert https://github.com/reasonml/reason-react/pull/875 

davesnx:
Only updating snapshots to run the CI, I could make the test enable_if based on OCAML version or maybe melange version (?) but still easier to keep main working


davesnx:
Both CIs can pass at the same time with different melange versions, that's a bit unfortunate, can we align on using same versions in CI or branch the CI per melange version? @anmonteiro

anmonteiro:
@davesnx I was thinking we should use the latest (5.0.1) but feel free to use the version you need to unblock you and i'll fix it for the nix build (feel free to merge a PR passing opam and failing nix)

PR #882, opened 2025-02-25
Install melange.5.0.0-52 in CI

davesnx:
- Force development to use melange 5 & ppxlib 0.33 (since we don't use a lock, I prefer to keep dev/test versions using `=`)
- Make tests using merlin more robust by only asserting what's visible on the editor
- Add annotations of broken tests
- Remove locations.t test since ppxlib's driver -dparsetree doesn't print locations when rendering the ast https://github.com/ocaml-ppx/ppxlib/pull/530

The first runs of the CI for this PR will fail, it's all on the schedule


ppx/test/keys.t/run.t
```diff
@@ -110,7 +110,7 @@ __^
 
 <img src={author.Author.imageUrl} />
 ______^
-
+TODO: This is a regression, the type is not correct should be a string
```
anmonteiro, line 113: I also caught this in one of my PRs. do you know if this is this related to the ppxlib upgrade? or have you not identified a cause yet?

davesnx: Didn't get deep into it, but want to revisit the locations of props/elements

davesnx: My plan now is to merge this, the React 19 and another PR investigating the locations

PR #887, opened 2025-04-03
Support melange.5.1.0-53

davesnx:


anmonteiro:
I updated the tests here, is  there anything missing before this can be merged?

test/melange-testing-library/react/ReactTestingLibrary.re
```diff
@@ -1,6 +1,8 @@
 open DomTestingLibrary;
 
-module FireEvent = FireEvent;
+module FireEvent = {
+  include FireEvent;
```
davesnx, line 4: is this change on purpose @anmonteiro ?

davesnx:
I want to make a release without ppxlib 0.36.0 (that would come with melange 5.1, if I'm not mistaken, per https://github.com/melange-re/melange/blob/main/melange.opam#L17) before merging this

PR #893, opened 2025-06-03
Prepare for 0.16.0

davesnx:
- Add unreleased PRs into CHANGES.md
- Fix melange dependency in dune-projects
- Update testing-library with React 19 version patch
- Remove version and other meta from package.json


davesnx:
I changed the demo to be only available in profile dev, which makes it only run in local and avoids CI/release etc

